PR #73, opened 2016-06-22
Need help debugging distinct property of undefined type 

felixcodes:
Incomplete / buggy code. 
Please see spec.ts
Please run npm run test and observe console.log output. console.log only outputs on failed cases. Fails on return statement below console.log in spec.ts.

Also: First compassQL failure was intentional.


kanitw:
@felixcodes  I assume you no longer need this PR?  If so, please close it.

PR #113, opened 2016-07-08
[WIP] scale_zero parameter

felixcodes:
I am not entirely sure if the constraint I added to encoding belongs in encoding or in spec. I also need to figure out what test and constraints go in spec, and need to write out enumerator test and other tests.


src/config.ts
```diff
@@ -93,6 +95,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxBinsList: [5, 10, 20],
   scaleTypes: [ScaleType.LINEAR, ScaleType.LOG],
+  scaleZeros: [true, false],
```
kanitw: Do you ever refer to this?  (I believe you use the default `[true, false]` [here](https://github.com/uwdata/compassql/pull/113/files#diff-a6322b30cb17d35f81779ff7980ea4c3R109))

src/constraint/encoding.ts
```diff
@@ -194,7 +194,22 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
-  }, {
+  },{
+    name: 'barMarkWithScaleZero',
+    description: 'Do not reccommend bar mark if scale does not start at zero',
+    properties: [Property.MARK, Property.SCALE_ZERO, Property.CHANNEL, Property.TYPE],
+    requireAllProperties: true,
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, stats: Stats, opt: QueryConfig) => {
+      if (contains([Channel.X, Channel.Y], encQ.channel) && (encQ.type === Type.QUANTITATIVE) &&
+          (encQ.type === Type.QUANTITATIVE) && encQ.scale && !encQ.bin) {
```
kanitw: This constraint have *many* issues.  

1) You have `mark` a part of the constraint.  Can it be encoding constraint?  For that reason, you haven't refer to mark at all.

2) why do you have to test `(encQ.type === Type.QUANTITATIVE)` twice?  

also why do you have to test for `encQ.bin` (which is probably irrelevant?)

If you really need to test bin here, you should add it to be a part of `properties`.  (Don't because it's irrrelvant -- unless you can prove that I overlook anything.)

src/property.ts
```diff
@@ -90,6 +94,11 @@ export const NESTED_ENCODING_PROPERTIES: NestedEncodingProperty[] = [
     property: Property.SCALE_TYPE,
     parent: 'scale',
     child: 'type'
+  },
+  {
+    property: Property.SCALE_ZERO,
```
kanitw: I would use `reduce` method to generate `NESTED_ENCODING_PROPERTIES` that would have the following output for now.  (Add a test!)

```
{
  bin: ['maxbins']
  scale: ['type', 'zero'] 
}
```

so that you can use it [here](https://github.com/uwdata/compassql/pull/113/files#diff-e723711227f2a0ee0232fccd5a5a1f35R273).  

Make sure that if you add another nested encoding property to 
`NESTED_ENCODING_PROPERTIES` you update this object for free.

src/query.ts
```diff
@@ -265,19 +267,24 @@ export function stringifyEncodingQueryFieldDef(encQ: EncodingQuery): string {
...
-      if (keys(scaleParams).length > 0) {
-        params.push({
-          key: 'scale',
-          value: JSON.stringify(scaleParams)
-        });
-      }
+
+      // // TODO: push other scale properties to scaleParams (add it into the array).
+
+        var scaleParams = ['type', 'zero']
```
kanitw: This looks correct.  However, we can reduce one location we have to update everytime we add a parameter by following this [suggestion](https://github.com/uwdata/compassql/pull/113/files#r70013361).

test/constraint/encoding.test.ts
```diff
@@ -295,6 +296,19 @@ describe('constraints/encoding', () => {
     });
   });
 
+  describe('barMarkWithScaleZero', () => {
```
kanitw: 1) See comment about the constraints and move this test accordingly.  
2) You test should **always** cover cases when it returns true and false.  Here you only add one case that cover when it's `false`.   Keep adding test until your constraint have full coverage.  (Use `npm run cover` and look at your constraint after you fix it.)

kanitw:
- [ ] `omitScaleZeroWithBinnedField`

- [ ] After you move it, rename `barMarkWithScaleZero` to `alwaysIncludeZeroInScaleWithBarMark`

- [ ] add `scaleZeroMustMatchScaleType`.    (Don’t use `zero` with scale types that does not support `zero` e.g., `log`, `ordinal` — We actually haven’t handled this well in Vega-Lite — so you might want to look at D3 scale docs (https://github.com/d3/d3-scale) to check if each type of scale supports zero.

kanitw:
- [ ] Since some of these constraints (`omitScaleZeroWithBinnedField`, `alwaysIncludeZeroInScaleWithBarMark`, `scaleZeroMustMatchScaleType`) might interact -- you might need to think about how to write an integrate test in generate.test that accounts for the interaction.  Basically, enumerate all relevant properties `mark`, `scale.zero`, `scale.type` and the output should still be correct.

src/property.ts
kanitw: Btw, rename `NESTED_ENCODING_PROPERTY_INDEX` to `NESTED_ENCODING_PROPERTY_PARENT_INDEX`

because the key is parent.  

(While `NESTED_ENCODING_INDEX` below, has the property itself as key.)

src/constraint/spec.ts
```diff
@@ -101,6 +101,49 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+    properties: [Property.MARK, Property.SCALE, Property.SCALE_ZERO, Property.CHANNEL, Property.TYPE],
+    requireAllPropertiesSpecific: true,
+    strict: true,
+    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      const mark = specM.getMark();
+      const encodings = specM.getEncodings();
+
+      if (mark === Mark.BAR) {
+        for (let encQ of encodings) {
+          if ((encQ.channel === Channel.X || encQ.channel === Channel.Y) && encQ.scale && encQ.type === Type.QUANTITATIVE) {
```
kanitw: This `if` is hard to read.  You might want to use contain and put predicates on 3-4 lines.

kanitw: Like
```
if( (predicate1) &&
     (predicate2) && ...
     ... ) {

File: src/constraint/spec.ts [hunk 7]
```diff
@@ -487,6 +530,27 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
       }
       return true;
     }
+  },
+  {
+    name: 'scaleZeroMustMatchScaleType',
+    description: 'ScaleZero should not be used with LOG, ORDINAL, TIME and UTC',
+    properties: [Property.SCALE, Property.SCALE_TYPE],
```
kanitw: missing scale zero

src/constraint/spec.ts
```diff
@@ -487,6 +530,27 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+    requireAllPropertiesSpecific: true,
+    strict: true,
+    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      const encodings = specM.getEncodings();
+
+      for (let encQ of encodings) {
+        if (encQ.scale) {
+          if (contains([ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC], (encQ.scale as ScaleQuery).type) &&
+             (encQ.scale as ScaleQuery).zero === true) {
```
kanitw: Just stylistic comment

You can assign `const scale: ScaleQuery = encQ.scale as ScaleQuery` so your predicates becomes shorter and easier to read.

src/constraint/spec.ts
felixcodes: I updated it to use the multiline && like you specified but it looks kind of funky to me. Please let me know if I did it correctly. Pushing in 30 seconds.

test/generate.test.ts
```diff
@@ -154,6 +154,73 @@ describe('generate', function () {
...
+              // type: {values: [undefined, ScaleType.SQRT, ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC]}
+            },
+            field: 'Q',
+            type: Type.QUANTITATIVE
+          }
+        ]
+      };
+      const answerSet = generate(specQ, schema);
+      assert.equal(answerSet.length, 1); // shouldn't this be 0? why is undefined passing?
+      assert.equal((answerSet[0].getEncodingQueryByIndex(0).scale as ScaleQuery).type, undefined);
```
felixcodes: @kanitw 
I do not understand why this is part of the answer set. I believe I wrote omitScaleZeroWithBinnedField correctly; the unit test is working as expected: if bin == true && zero == true, the constraint fails.

However in my generate test, this one case with undefined seems to be getting through. I'm not sure what I'm misunderstanding.

src/query.ts
```diff
@@ -312,19 +315,22 @@ export function stringifyEncodingQueryFieldDef(encQ: EncodingQuery): string {
...
-      if (keys(scaleParams).length > 0) {
-        params.push({
-          key: 'scale',
-          value: JSON.stringify(scaleParams)
-        });
-      }
+
+        const scaleParamsArr = getNestedEncodingPropertyChild('scale');
+        var scaleParams = scaleParamsArr.reduce((scaleParamsObj, param) => {
+            console.log(param);
```
kanitw: @felixcodes  Please remove all console.log

src/property.ts
```diff
@@ -129,6 +153,10 @@ export function getNestedEncodingProperty(prop: Property) {
   return NESTED_ENCODING_INDEX[prop];
 }
 
+export function getNestedEncodingPropertyChild(parent: string) {
```
kanitw: is parent an arbritary string?  (No!  It should be `parent: Property`)

test/generate.test.ts
```diff
@@ -154,6 +154,98 @@ describe('generate', function () {
...
+            field: 'Q',
+            type: Type.QUANTITATIVE
+          }
+        ]
+      };
+      const answerSet = generate(specQ, schema);
+
+      /* note for future developer:
+        You might expect this answerset to be completely empty since there is a constraint that prevents bin and zero from working together, however
+        if you look inside generate() you'll see that the checks for bin/zero only occur if they're enumerated. Since they're true/false values they don't run through
```
kanitw: "the checks for bin/zero only occur"  => "we only call <constraint name>, which requires Property.BIN and Property.SCALE_ZERO, if at least one of the properties is enumerated.  

"true/false" ==> I don't see any false here.  Just "true"

Can you be more careful with writing comments?  Otherwise your comments just add more confusion to readers.

test/query.test.ts
```diff
@@ -467,6 +467,13 @@ describe('query', () => {
        assert.equal(str, 'a,q,scale={"type":"log"}');
     });
 
+    it('should return correct fieldDef string for scale with scaleZero', () => {
```
kanitw: with zero=true

src/query.ts
```diff
@@ -312,19 +315,22 @@ export function stringifyEncodingQueryFieldDef(encQ: EncodingQuery): string {
...
+            if (encQ.scale[param as string]) {
```
kanitw: Stylistically, I don't think you need any `param as string` here -- it reads quite annoyingly long.  If you really do due to error, make `param: any` above in the method signature.

test/generate.test.ts
```diff
@@ -216,9 +216,9 @@ describe('generate', function () {
       const answerSet = generate(specQ, schema);
 
       /* note for future developer:
-        You might expect this answerset to be completely empty since there is a constraint that prevents bin and zero from working together, however
-        if you look inside generate() you'll see that the checks for bin/zero only occur if they're enumerated. Since they're true/false values they don't run through
-        that check and undefined is able to pass through.
+        You might expect this answerset to be completely empty since there is a constraint that prevents bin and zero from working together,
+        however if you look inside generate() you'll see that we only call, which requires Property.BIN and Property.SCALE_ZERO,
```
kanitw: we only call what?  ..........

felixcodes: omitScaleZeroWithBinnedField

PR #117, opened 2016-07-08
Add summary stats to FieldSchema

RileyChang:
1. Schema now stores a data `Summary` in each of its `FieldSchema`'s
2. Schema has a `stats` method for retrieving the `Summary`


kanitw:
Still missing test, but I'm merging for now so I can use this for refactoring.   Please send a follow up for adding test :)

PR #119, opened 2016-07-08
Add optional config to `query()`

kanitw:
(from slack conversation)


kanitw:
I'll merge this for now.

@RileyChang  Please send follow up comment if you have any concern with this. 

In general, try to prioritize doing code review before continue writing code so you won't block others :)

PR #120, opened 2016-07-08
Low cardinality numerical data is inferred to be ordinal

### LINKED ISSUE #88: Set number field to ordinal if cardinality is low
- [ ] For `number` and `integer` field, we currently always set them to `quantitative` (in vega-editor).  However, if the cardinality of an `integer` field is low (~~< some constant in config e.g., `8`~~ -- see below for better idea) we might want to consider them as `ordinal` or `nominal`.
(If cardinality is N and the numbers are 1,2,3, N, we can set it to `nominal`. Otherwise set to `ordinal`)

RileyChang:
Fix #88

1. Fields with integer primitive type are inferred to be ordinal if (distinct / total count) is less than a given percentage.
2. `Schema.build` takes in an optional `QueryConfig` options object.
3. Default percentage is specified in config.ts (5%)
4. Wrote tests for behavior in (1)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,7 +30,7 @@ export interface QueryConfig {
   /** Default types to enumerate */
   types?: Type[];
 
-  /** Default ratio for numbers to be considered ordinal */
+  /** Default ratio for number fields to be considered ordinal */
   numberOrdinalProportion?: number;
 
   /** Default maxbins to enumerate */
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,6 +30,9 @@ export interface QueryConfig {
   /** Default types to enumerate */
   types?: Type[];
 
+  /** Default ratio for numbers to be considered ordinal */
+  numberOrdinalProportion?: number;
+
   /** Default maxbins to enumerate */
   maxBinsList?: number[];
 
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -30,6 +30,9 @@ export interface QueryConfig {
   /** Default types to enumerate */
   types?: Type[];
 
+  /** Default ratio for numbers to be considered ordinal */
```
kanitw: "for number fields" ?

src/schema.ts
```diff
@@ -14,17 +16,39 @@ export class Schema {
...
+      var type: Type;
+      if (primitiveType === PrimitiveType.NUMBER) {
+        type = Type.QUANTITATIVE;
+      } else if (primitiveType === PrimitiveType.INTEGER) {
+        // use ordinal or nominal when cardinality of integer type is relatively low
+        if (distinct / summary.count < opt.numberOrdinalProportion) {
+          // use nominal if the integers are 1,2,3,...,N where N = cardinality
+          var dataPoints: number[] = keys(summary.unique) as any;
+          var min = dataPoints[0], max = dataPoints[0];
+          for (var i = 1; i < dataPoints.length; i++) {
```
kanitw: why don't you sort the array and check them one by one?

kanitw: I guess you want to support 0, ..., N-1 case too?  

In any case, if you don't want to check each of them one-by-one and wanna use only `min` and `max` -- we already have `summary.min` and `summary.max` so this for-loop seem unnecessary.

kanitw:
Given that I add more check for min in https://github.com/uwdata/compassql/pull/120/commits/5f668d5164646356dd0c6ac8e4f7bafa5b310fcc, you might add a few more tests and I think we're good go to merge this :)

PR #123, opened 2016-07-09
[WIP] Update both constraints to support nested encoding property

felixcodes:
Haven't written any tests for Scale_Type and Scale_Zero yet. 
Just want to make sure I'm on the right track :) 
@kanitw

### DIFF
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -3,7 +3,7 @@ import {Type} from 'vega-lite/src/type';
 
 import {QueryConfig} from '../config';
 import {EnumSpecIndexTuple, SpecQueryModel} from '../model';
-import {Property} from '../property';
+import {getNestedEncodingProperty, hasNestedProperty, Property} from '../property';
 import {EncodingQuery, isEnumSpec, isDimension, isMeasure, ScaleQuery} from '../query';
 import {PrimitiveType, Schema} from '../schema';
 import {contains, some} from '../util';
@@ -31,7 +31,16 @@ export class EncodingConstraintModel extends AbstractConstraintModel {
       // TODO: extract as a method and do unit test
       const hasRequiredPropertyAsEnumSpec = some(
         this.constraint.properties,
-        (prop) => isEnumSpec(encQ[prop])
+        function(prop) {
+          if (hasNestedProperty(prop)) {
+            let nestedEncProp = getNestedEncodingProperty(prop);
+            let parent = nestedEncProp.parent;
+            let child = nestedEncProp.child;
+            return isEnumSpec(encQ[parent][child]);
+          } else {
+            return isEnumSpec(encQ[prop]);
+          }
+        }
       );
       // If one of the required property is still an enum spec, do not check the constraint yet.
       if (hasRequiredPropertyAsEnumSpec) {
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -3,10 +3,10 @@ import {Type} from 'vega-lite/src/type';
 
 import {QueryConfig} from '../config';
 import {EnumSpecIndexTuple, SpecQueryModel} from '../model';
-import {getNestedEncodingProperty, hasNestedProperty, Property} from '../property';
+import {getNestedEncodingProperty, Property} from '../property';
 import {EncodingQuery, isEnumSpec, isDimension, isMeasure, ScaleQuery} from '../query';
 import {PrimitiveType, Schema} from '../schema';
-import {contains, some, every} from '../util';
+import {contains, every} from '../util';
 import {ScaleType} from 'vega-lite/src/scale';
 
 import {AbstractConstraint, AbstractConstraintModel} from './base';
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -6,7 +6,7 @@ import {EnumSpecIndexTuple, SpecQueryModel} from '../model';
 import {getNestedEncodingProperty, hasNestedProperty, Property} from '../property';
 import {EncodingQuery, isEnumSpec, isDimension, isMeasure, ScaleQuery} from '../query';
 import {PrimitiveType, Schema} from '../schema';
-import {contains, some} from '../util';
+import {contains, some, every} from '../util';
 import {ScaleType} from 'vega-lite/src/scale';
 
 import {AbstractConstraint, AbstractConstraintModel} from './base';
@@ -25,27 +25,32 @@ export class EncodingConstraintModel extends AbstractConstraintModel {
     super(constraint);
   }
 
+  public hasAllRequiredProperties(encQ: EncodingQuery): boolean {
+    return every(this.constraint.properties, (prop) => {
+      const nestedEncProp = getNestedEncodingProperty(prop);
+
+      if (nestedEncProp) {
+        let parent = nestedEncProp.parent;
+        let child = nestedEncProp.child;
+
+        if (!encQ[parent]) {
+          return true;
+        }
+
+        return !isEnumSpec(encQ[parent][child]);
+      }
+
+      return !isEnumSpec(encQ[prop]);
+    });
+  }
+
   public satisfy(encQ: EncodingQuery, schema: Schema, opt: QueryConfig): boolean {
     // TODO: Re-order logic to optimize the "requireAllProperties" check
     if (this.constraint.requireAllProperties) {
       // TODO: extract as a method and do unit test
-      const hasRequiredPropertyAsEnumSpec = some(
-        this.constraint.properties,
-        function(prop) {
-          const nestedEncProp = getNestedEncodingProperty(prop);
-          if (nestedEncProp) {
-            if (nestedEncProp.parent === prop) {
-              let child = nestedEncProp.child;
-              return isEnumSpec(encQ[prop][child]);
-            }
-          } else {
-            return isEnumSpec(encQ[prop]);
-          }
-        }
-      );
-      // If one of the required property is still an enum spec, do not check the constraint yet.
-      if (hasRequiredPropertyAsEnumSpec) {
-        return true; // Return true since the query still satisfy the constraint.
+
+      if (!this.hasAllRequiredProperties(encQ)) {
+        return true;
       }
     }
     return (this.constraint as EncodingConstraint).satisfy(encQ, schema, opt);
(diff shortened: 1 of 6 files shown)


src/constraint/encoding.ts
```diff
@@ -31,7 +31,16 @@ export class EncodingConstraintModel extends AbstractConstraintModel {
       // TODO: extract as a method and do unit test
       const hasRequiredPropertyAsEnumSpec = some(
         this.constraint.properties,
-        (prop) => isEnumSpec(encQ[prop])
+        function(prop) {
+          if (hasNestedProperty(prop)) {
+            let nestedEncProp = getNestedEncodingProperty(prop);
```
kanitw: Please use `const` instead of `let` because you never modify these values.

src/constraint/encoding.ts
```diff
@@ -31,7 +31,16 @@ export class EncodingConstraintModel extends AbstractConstraintModel {
...
+            let parent = nestedEncProp.parent;
+            let child = nestedEncProp.child;
+            return isEnumSpec(encQ[parent][child]);
+          } else {
```
kanitw: You `if` always return.  So you can actually remove `else` to have less text.  
(just a style comment -- no big deal.  But I prefer that you remove :) )

src/constraint/spec.ts
```diff
@@ -51,6 +51,18 @@ export class SpecConstraintModel extends AbstractConstraintModel {
               return some(specM.getEncodings(), (encQ) => {
                 return isEnumSpec(encQ[prop]);
               });
+
+            case Property.SCALE_TYPE:
```
kanitw: Inspired from `encoding.ts` you can actually refactor to replace this `switch` with `if-else` and avoid having to add new property to the switch every time. 


Hint:  
- if mark
- leave // TODO transform there
- use `ENCODING_PROPERTIES` for second group
- `getNestedEncodingProperty` like how you suppose to use it in `encoding.ts`.   (Of course you have to figure that out first)

src/constraint/encoding.ts
```diff
@@ -32,11 +32,12 @@ export class EncodingConstraintModel extends AbstractConstraintModel {
...
         this.constraint.properties,
         function(prop) {
-          if (hasNestedProperty(prop)) {
-            let nestedEncProp = getNestedEncodingProperty(prop);
-            let parent = nestedEncProp.parent;
-            let child = nestedEncProp.child;
-            return isEnumSpec(encQ[parent][child]);
+          const nestedEncProp = getNestedEncodingProperty(prop);
+          if (nestedEncProp) {
+            if (nestedEncProp.parent === prop) {
```
kanitw: Why should `parent === prop`?    Have you really use to debugger to test this?

Look at what are values of `prop`, `parent`, `child`.

kanitw:
- [ ] Obviously you haven't add the missing `Property.SCALE_TYPE` from relevant constraints.

src/constraint/spec.ts
```diff
@@ -23,45 +23,42 @@ export class SpecConstraintModel extends AbstractConstraintModel {
     super(specConstraint);
   }
 
+    public hasAllRequiredProperties(specM: SpecQueryModel): boolean {
+      return every(this.constraint.properties, (prop) => {
+        if (prop === Property.MARK) {
+          return !isEnumSpec(specM.getMark());
+        }
+
```
kanitw: Add // TODO: transform here.

kanitw: Also should still throw error if prop is not mark and not encoding property to warn us to implement the right behavior here.

src/constraint/spec.ts
```diff
@@ -23,45 +23,50 @@ export class SpecConstraintModel extends AbstractConstraintModel {
...
+            return !isEnumSpec(encQ[parent][child]);
+          });
         }
-      );
-      // If one of the required property is still an enum spec, do not check the constraint yet.
-      if (hasRequiredPropertyAsEnumSpec) {
-        return true; // Return true since the query still satisfy the constraint.
+
+        if (!contains(ENCODING_PROPERTIES, prop)) {
```
kanitw: This takes `O(N)` to just check for error every time and this method get called very often.  Do you think this is an appropriate implementation?

src/constraint/spec.ts
kanitw: You haven't added this TODO, although it's still left over here in the PR.

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
     }
   });
 
+  describe('hasAllRequiredPropertiesSpecific', () => {
+    let encModel = new EncodingConstraintModel(
+      {
+        name: 'Test EncodingModel for hasAllRequiredProperties class method',
```
kanitw: "TestEncoding"

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
...
+        description: 'Test EncodingModel for hasAllRequirdProperties class method',
+        properties: [Property.AGGREGATE, Property.TYPE, Property.SCALE, Property.SCALE_TYPE],
+        requireAllPropertiesSpecific: true,
+        strict: true,
+        satisfy: undefined
+      }
+    );
+
+    let encQ1: EncodingQuery = {
```
kanitw: You never re-use this encQ1 for multiple test cases.  Why don't you put it inside relevant `it()` so that you can name it `encQ` without `1`.   (Same for all others)

test/constraint/spec.test.ts
```diff
@@ -32,6 +32,63 @@ describe('constraints/spec', () => {
...
+        name: 'Test SpecModel for hasAllRequiredPropertiesSpecific class method',
+        description: 'Test SpecModel for hasAllRequiredPropertiesSpecific class method',
+        properties: [Property.AGGREGATE, Property.TYPE, Property.SCALE, Property.SCALE_TYPE, Property.MARK],
+        requireAllPropertiesSpecific: true,
+        strict: true,
+        satisfy: undefined
+      }
+    );
+
+    const specQM1 = buildSpecQueryModel({
```
kanitw: Put inside `it`, remove `1` suffix.  Same for all others.

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
...
+      channel: Channel.X,
+      aggregate: AggregateOp.MEAN,
+      field: 'A',
+      scale: {type: ScaleType.LOG},
+      type: Type.QUANTITATIVE
+    };
+
+    it('should return true if all properties are present', () => {
```
kanitw: "are present" => "is defined"

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
...
+    });
+
+    let encQ2: EncodingQuery = {
+      channel: Channel.X,
+      field: 'A',
+      scale: {type: ScaleType.LOG},
+      type: Type.QUANTITATIVE
+    };
+
+    it('should return true if a required property is not present', () => {
```
kanitw: "not present" => "undefined"

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
...
+    });
+
+    let encQ3: EncodingQuery = {
+      channel: Channel.X,
+      aggregate: SHORT_ENUM_SPEC,
+      scale: {type: ScaleType.LOG},
+      type: Type.QUANTITATIVE
+    };
+
+    it('should return false if a required property is enum spec', () => {
```
kanitw: an enum spec

test/constraint/encoding.test.ts
```diff
@@ -26,6 +27,65 @@ describe('constraints/encoding', () => {
...
+
+    let encQ4: EncodingQuery = {
+      channel: Channel.X,
+      aggregate: AggregateOp.MEAN,
+      field: 'A',
+      scale: {type: SHORT_ENUM_SPEC},
+      type: Type.QUANTITATIVE
+    };
+
+    it('should return false if a nested required property is enum spec', () => {
```
kanitw: an enum spec

test/constraint/spec.test.ts
```diff
@@ -32,6 +32,63 @@ describe('constraints/spec', () => {
...
+      mark: Mark.POINT,
+      encodings: [
+        {aggregate: AggregateOp.MEAN, channel: Channel.X, field: 'A', scale: {type: ScaleType.LOG}, type: Type.QUANTITATIVE}
+      ]
+    });
+
+    it('should return true if all properties are present', () => {
```
kanitw: is defined

test/constraint/spec.test.ts
```diff
@@ -32,6 +32,63 @@ describe('constraints/spec', () => {
...
+    });
+
+    const specQM2 = buildSpecQueryModel({
+      mark: Mark.POINT,
+      encodings: [
+        {aggregate: AggregateOp.MEAN, channel: Channel.X, field: 'A',  type: Type.QUANTITATIVE}
+      ]
+    });
+
+    it('should return true if a required property is not present', () => {
```
kanitw: is undefined

test/constraint/spec.test.ts
```diff
@@ -32,6 +32,63 @@ describe('constraints/spec', () => {
...
+    });
+
+    const specQM3 = buildSpecQueryModel({
+      mark: Mark.POINT,
+      encodings: [
+        {aggregate: AggregateOp.MEAN, channel: Channel.X, field: 'A', scale: SHORT_ENUM_SPEC, type: Type.QUANTITATIVE}
+      ]
+    });
+
+    it('should return false if a required property is EnumSpec', () => {
```
kanitw: an enum spec

test/constraint/spec.test.ts
```diff
@@ -32,6 +32,63 @@ describe('constraints/spec', () => {
...
+    });
+
+    const specQM4 = buildSpecQueryModel({
+      mark: Mark.POINT,
+      encodings: [
+        {aggregate: AggregateOp.MEAN, channel: Channel.X, field: 'A', scale: {type: SHORT_ENUM_SPEC}, type: Type.QUANTITATIVE}
+      ]
+    });
+
+    it('should return false if a nested required property is EnumSpec', () => {
```
kanitw: an enum spec

src/property.ts
```diff
@@ -32,6 +36,8 @@ export function hasNestedProperty(prop: Property) {
     case Property.TYPE:
     case Property.BIN_MAXBINS:
     case Property.SCALE_TYPE:
+    case Property.TRANSFORM_CALCULATE:
```
kanitw: I don't know why you have to add TRANSFORM_CALCULATE now.  

`TRANSFORM_CALCULATE` clearly have nested property  (`field` and `expr`) ?

kanitw: I would say remove all of transform for now.  That's not the point of this issue and you're not doing it correctly anyway.

PR #124, opened 2016-07-09
Refactor API

kanitw:
- `query()` should return augmented `query` in addition to the `result: SpecQueryModelGroup`.
-  config inside `query()` instead of generate for consistent config
- No longer worried about causing side effect in `SpecQueryModel.build()` as we only have to prevent side effect to the original `q: Query`
- make required `duplicate()` in `query.normalize()`

- Fix test + move leftover generate test in `enumerate.test.ts` to `generate.test.ts`

### DIFF
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -4,11 +4,8 @@ import {QueryConfig, DEFAULT_QUERY_CONFIG} from './config';
 import {SpecQueryModel} from './model';
 import {SpecQuery} from './query';
 import {Schema} from './schema';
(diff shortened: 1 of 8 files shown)


kanitw:
@felixcodes  @RileyChang  I'm gonna just merge this, but you should be aware of this change & feel free to send follow-up comment.  

(See https://github.com/uwdata/compassql/commit/21aa2554e6a767e14094e7943afa4f00424492a9 for pure changes.  -- I split between actual change and test to facilitate reading.)

Issue #130, opened 2016-07-13
Rename `requireAllPropertiesSpecific` to `allowEnumSpecForProperties`


kanitw:
`requireAllPropertiesSpecific` still sound vague? 

I propose that we replace

`requireAllPropertiesSpecific:  true` with `allowEnumSpecForProperties: false` + vice versa
and  replace `if (requireAllPropertiesSpecific)`  with `if (!allowEnumSpecForProperties)` ... something like this? 

What do you think?

felixcodes:
Will probably do that after all of the scale stuff is finished. 

reminder for future sleepy-felix: this means flipping the logic in some expressions as well...

PR #133, opened 2016-07-14
RequireAllPropertiesSpecific --> allowEnumSpecForProperties

felixcodes:
Refactor for #130.

RequireAllPropertiesSpecific = true -->
allowEnumSpecForProperties = false

and vice versa


kanitw:
@felixcodes   Can you say "Fix #130" next time if your PR address all points in the issue, so we don't have to manualy close #130.   (If your PR only address a part of an issue, don't say "fix".)

PR #134, opened 2016-07-14
[WIP] Scale bandSize

felixcodes:
Part of #2. 

Add basic support for scale.clamp, scale.exponent, scale.round. Add scaleType() helper method used for inferring correct scale type from encoding's type. Includes tests for scale properties and scaleType helper.


felixcodes:
@kanitw Scale bandSize is no longer WIP.

kanitw:
Test is failing.

src/config.ts
```diff
@@ -98,6 +100,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   numberOrdinalProportion: .05,
 
   maxBinsList: [5, 10, 20],
+  scaleBandSizes: [0, 21, 90],
```
kanitw: why 90?

src/constraint/spec.ts
```diff
@@ -76,6 +78,26 @@ export interface SpecConstraint extends AbstractConstraint {
   satisfy: SpecConstraintChecker;
 }
 
+export function getTrueScaleType(encQ: EncodingQuery): ScaleType {
```
kanitw: rename to just `scaleType`

kanitw: Make sure to add unit test for this.

kanitw: This should be in `query.ts` and soon move to `query/encoding.ts`

src/constraint/spec.ts
```diff
@@ -532,9 +554,30 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+     name: 'scaleBandSizeMustMatchScaleType',
```
kanitw: Our last conversation, we mentiont that we only want one `scalePropertiesSupportedByScaleType` rather than having one for each scale property.   Please jot down something we already agree on.  It's not fun to have to keep repeating something for you.   

Plus, why is this a spec constraint?  (It seems like an encoding constraint for me.)

src/config.ts
kanitw: 0 is totally unacceptable default value for this `scale.bandSize`. 

Why on the earth would people want a band size to be zero?

src/query.ts
```diff
@@ -345,3 +346,23 @@ export function stringifyEncodingQueryFieldDef(encQ: EncodingQuery): string {
...
+    } else if (type === Type.ORDINAL || type === Type.NOMINAL) {
+      return ScaleType.ORDINAL;
+    } else if (type === Type.TEMPORAL) {
+      if (encQ.timeUnit) {
+        return defaultScaleType(encQ.timeUnit as TimeUnit);
+      } else {
+        return ScaleType.TIME;
+      }
+    }
+  }
```
kanitw: This method does not always return a value.  Does it pass compilation at all?

src/property.ts
```diff
@@ -160,3 +170,110 @@ export function getNestedEncodingPropertyChild(parent: Property) {
 export function isNestedEncodingProperty(prop: Property) {
   return prop in NESTED_ENCODING_INDEX;
 }
+
+const SUPPORTED_SCALE_PROPERTY =
```
kanitw: Why is this added in this PR?  Do you use it at all?

src/query.ts
```diff
@@ -345,3 +346,23 @@ export function stringifyEncodingQueryFieldDef(encQ: EncodingQuery): string {
     params.map((p) => ',' + p.key + '=' + p.value).join('');
   return (fn ? fn + '(' + fieldType + ')' : fieldType);
 }
+
+export function scaleType(encQ: EncodingQuery): ScaleType {
```
kanitw: Don't take the whole `encQ` as input. Take `scaleType`, `timeUnit`, and `type` as inputs so when you call it from constraint, it's clear that you only use only these properties of `encQ`.

src/constraint/encoding.ts
```diff
@@ -221,7 +221,7 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: true,
     satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
       if (encQ.scale) {
-        const scaleType = (encQ.scale as ScaleQuery).type;
+        const scaleType = actualScaleType(encQ);
```
kanitw: I don't like `actualScaleType` ... maybe name the variable `_scaleType` ?

src/constraint/spec.ts
```diff
@@ -323,7 +324,14 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     if (mark === Mark.AREA || mark === Mark.BAR) {
       for (let encQ of encodings) {
         if((encQ.channel === Channel.X || encQ.channel === Channel.Y) && encQ.scale) {
-          if (((encQ.scale as ScaleQuery).type as ScaleType) === ScaleType.LOG) {
+
+          const scale: ScaleQuery = encQ.scale as ScaleQuery;
+          let sType = scale.type as ScaleType;
```
kanitw: I don't think as need declarations for `sType, timeUnit, type` here as we only use them once.  Just pass them to the method.

src/constraint/spec.ts
```diff
@@ -323,7 +324,14 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+          let timeUnit = encQ.timeUnit as TimeUnit;
+          let type = encQ.type as Type;
+          let scaleType = _scaleType(sType, timeUnit, type);
```
kanitw: Umm, please read my comments carefully.   I mentioned that "use `_scaleType` as local variable and keep method name `scaleType`.

(That said, `sType` maybe better than `_scaleType`.)

src/constraint/spec.ts
```diff
@@ -534,16 +542,22 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
     strict: true,
     satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
       const encodings = specM.getEncodings();
 
       for (let encQ of encodings) {
         if (encQ.scale) {
+
           const scale: ScaleQuery = encQ.scale as ScaleQuery;
-          if (contains([ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC], scale.type) &&
+          let sType = scale.type as ScaleType;
```
kanitw: unnecessary declaration

src/query/encoding.ts
```diff
@@ -48,3 +48,22 @@ export function isMeasure(encQ: EncodingQuery) {
...
+  if (scaleType !== undefined) {
+    return scaleType;
+  }
+
+  if (type === Type.QUANTITATIVE) {
+    return ScaleType.LINEAR;
+  } else if (type === Type.ORDINAL || type === Type.NOMINAL) {
+    return ScaleType.ORDINAL;
+  } else {
+    // Down here, type is guaranteeed to be Type.TEMPORAL
```
kanitw: > type is guaranteeed to be Type.TEMPORAL

How do you assume this?
  
This is not true.  Type can be enumSpec in this case.  Not to mention that we might support more type in the future.   

Check for temporal, check for enum spec and throw error for else case.

src/query/encoding.ts
```diff
@@ -48,3 +48,22 @@ export function isMeasure(encQ: EncodingQuery) {
...
+    if (timeUnit !== undefined) {
+      return defaultScaleType(timeUnit);
```
kanitw: Does this behave correctly with enum spec?

src/query/encoding.ts
```diff
@@ -48,3 +48,22 @@ export function isMeasure(encQ: EncodingQuery) {
   return (encQ.type === Type.QUANTITATIVE && !encQ.bin) ||
       (encQ.type === Type.TEMPORAL && !encQ.timeUnit);
 }
+
+export function scaleType(scaleType: ScaleType, timeUnit: TimeUnit, type: Type) {
```
kanitw: This method doesn't have unit test yet.

src/query/encoding.ts
kanitw: (If so, make sure to have 100% test coverage for this method.)

src/constraint/encoding.ts
```diff
@@ -221,8 +222,10 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: true,
     satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
       if (encQ.scale) {
-        const scaleType = (encQ.scale as ScaleQuery).type;
-        const type = encQ.type;
+        const sType = (encQ.scale as ScaleQuery).type as ScaleType;
+        const timeUnit = encQ.timeUnit as TimeUnit;
+        const type = encQ.type as Type;
+        const scaleType = _scaleType(sType, timeUnit, type);
```
kanitw: 1) Unnecessary declarations, why don't you pass these parameters directly to the methods?
2) I commented that use _scaleType as local variable (In hindsight, `sType` might be better) and keep the method called `scaleType`.

src/query/encoding.ts
```diff
@@ -48,3 +48,38 @@ export function isMeasure(encQ: EncodingQuery) {
...
       (encQ.type === Type.TEMPORAL && !encQ.timeUnit);
 }
+
+/**
+ *  Returns the true scale type of an encoding.
+ *  @returns {ScaleType} If the scale type was not specified, it is inferred from the encoding's Type.
+ *  @returns {undefined} If the scale type was not specified and Type (or TimeUnit if applicable) is an EnumSpec, there is no clear scale type
+ */
+
+export function scaleType(scaleType: ScaleType, timeUnit: TimeUnit | EnumSpec<TimeUnit> | ShortEnumSpec, type: Type | EnumSpec<Type> | ShortEnumSpec) {
```
kanitw: Why the other two parameters can be enum spec but scaleType cannot be enum spec? 

I wonder if you can actually simplify this method to never accept enum spec? 
(Im not sure about that you need to check all reference if you always have them in specific form.)

src/constraint/spec.ts
```diff
@@ -323,7 +324,12 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     if (mark === Mark.AREA || mark === Mark.BAR) {
       for (let encQ of encodings) {
         if((encQ.channel === Channel.X || encQ.channel === Channel.Y) && encQ.scale) {
-          if (((encQ.scale as ScaleQuery).type as ScaleType) === ScaleType.LOG) {
+
+          let sType = scaleType((encQ.scale as ScaleQuery).type as ScaleType,
+                                    encQ.timeUnit as TimeUnit,
```
kanitw: tabbing not aligned.

src/query/encoding.ts
```diff
@@ -48,3 +48,38 @@ export function isMeasure(encQ: EncodingQuery) {
...
+    if (timeUnit !== undefined) {
+      if (isEnumSpec(timeUnit)) {
+        return undefined;
+      }
+      return defaultScaleType(timeUnit as TimeUnit);
+    } else {
+      return ScaleType.TIME;
+    }
+  } else {
+    /* istanbul ignore next */
```
kanitw: Is this truly the right location for ignore next?  Does test coverage say that this entire else branch is ignored?   (I doubt it should be one line above.)

src/constraint/spec.ts
```diff
@@ -543,8 +549,10 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
       for (let encQ of encodings) {
         if (encQ.scale) {
           const scale: ScaleQuery = encQ.scale as ScaleQuery;
-          if (contains([ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC], scale.type) &&
-             (scale.zero === true)) {
+
+          if (contains([ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC],
+              scaleType(scale.type as ScaleType, encQ.timeUnit as TimeUnit, encQ.type as Type)) &&
```
kanitw: assign `const sType = ...`   This line is getting too long.  (But no need for declaration for each parameter.)

src/model.ts
```diff
@@ -54,6 +54,9 @@ export interface EnumSpecIndex {
   /** List of indice tuples of encoding mappings that require scale enumeration. */
   scale?: EnumSpecIndexTuple<boolean>[];
 
+  /** List of indice tuple for encoding mappings that require enumerating scale.scale_bandsize */
```
kanitw: scale.bandSize

src/model.ts
```diff
@@ -54,6 +54,9 @@ export interface EnumSpecIndex {
...
+  scaleBandSize?: EnumSpecIndexTuple<number>[];
+
   /** List of indice tuple for encoding mappings that require enumerating scale.scale_type */
```
kanitw: scale.type

src/query/encoding.ts
kanitw: Thinking more about it, just make it accept enum spec so you can eliminate all lengthy `as` casting in all of usage of this function.

src/constraint/encoding.ts
```diff
@@ -222,24 +221,22 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: true,
     satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
       if (encQ.scale) {
-        const sType = (encQ.scale as ScaleQuery).type as ScaleType;
-        const timeUnit = encQ.timeUnit as TimeUnit;
         const type = encQ.type as Type;
-        const scaleType = _scaleType(sType, timeUnit, type);
+        const sType = scaleType((encQ.scale as ScaleQuery).type, encQ.timeUnit, encQ.type as Type);
```
kanitw: why still `as Type` here?

src/constraint/spec.ts
```diff
@@ -326,8 +326,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
         if((encQ.channel === Channel.X || encQ.channel === Channel.Y) && encQ.scale) {
 
           let sType = scaleType((encQ.scale as ScaleQuery).type as ScaleType,
```
kanitw: why still using these as since you already correct `scaleType` 's signature?

test/query/encoding.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      const sType = scaleType(undefined, SHORT_ENUM_SPEC, Type.TEMPORAL);
+      assert.equal(sType, undefined);
+    });
+
+    it('should return ScaleType.TIME if type is temporal and scale type and TimeUnit are not specified', () => {
+      const sType = scaleType(undefined, undefined, Type.TEMPORAL);
+      assert.equal(sType, ScaleType.TIME);
+    });
+
+    [TimeUnit.HOURS, TimeUnit.DAY, TimeUnit.MONTH, TimeUnit.QUARTER].forEach((timeU) => {
```
kanitw: why `timeU`  -- we use `timeUnit` everywhere

test/query/encoding.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      it('should return ScaleType.ORDINAL if type is temporal and TimeUnit is ' + timeU + ' and scale type is not specified', () => {
+        const sType = scaleType(undefined, timeU, Type.TEMPORAL);
+        assert.equal(sType, ScaleType.ORDINAL);
+      });
+    });
+
+    [TimeUnit.YEAR, TimeUnit.DATE, TimeUnit.MINUTES, TimeUnit.SECONDS, TimeUnit.MILLISECONDS, TimeUnit.YEARMONTH,
+     TimeUnit.YEARMONTHDATE, TimeUnit.YEARMONTHDATEHOURS, TimeUnit.YEARMONTHDATEHOURSMINUTES, TimeUnit.YEARMONTHDATEHOURSMINUTESSECONDS,
+     TimeUnit.HOURSMINUTES, TimeUnit.HOURSMINUTESSECONDS, TimeUnit.MINUTESSECONDS, TimeUnit.SECONDSMILLISECONDS,
+     TimeUnit.YEARQUARTER, TimeUnit.QUARTERMONTH, TimeUnit.YEARQUARTERMONTH].forEach((timeU) => {
```
kanitw: timeU....

src/constraint/spec.ts
```diff
@@ -325,8 +325,8 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
       for (let encQ of encodings) {
         if((encQ.channel === Channel.X || encQ.channel === Channel.Y) && encQ.scale) {
 
-          let sType = scaleType((encQ.scale as ScaleQuery).type as ScaleType,
-                                 encQ.timeUnit as TimeUnit, encQ.type as Type);
+          let sType = scaleType((encQ.scale as ScaleQuery).type,
```
kanitw: Why still leaving `as TimeUnit`?

Issue #22, opened 2016-06-01
Criteria for enumerating timeUnit


kanitw:
- [ ] We need good criteria for determine timeUnits that we should determine by default for a particular dataset 

Blocked #95  -- need stats first

kanitw:
@RileyChang   Basically, if a particular timeUnit gives cardinality `distinct` of 1, then it is a bad timeUnit to enumerate (because you don't see variation of `timeUnit` at all!) so we should prune it from the result list.   

This should be a pretty simple encoding constraint that takes `field` and `timeUnit` into account.
(`properties: [Property.FIELD, Property.TIMEUNIT], requireAllPropertiesSpecific: true`) and reads cardinality from the `schema` and return false if a particular timeUnit has cardinality ≤ 1.

See `src/constraint/encoding.ts` for example of other constraint.  @felixcodes is pretty knowledgable about constraint, so he could help if you get confused.  

Note we might rename `requireAllPropertiesSpecific` -- see #130

kanitw:
- [ ] From #109, since `schema` can be optional, check if schema is defined, if not, return true. (Because we can't know if the spec violates the constraint, so let it pass. We only prune surely bad things.)

PR #135, opened 2016-07-14
Cardinality functionality for bin/timeUnit fields - Fix #95

### LINKED ISSUE #95: Add Stats Calculation for binned fields and fields with timeUnit
Currently we don't have stats e.g., cardinality (`distinct`) for `bin`ned field, fields with `timeUnit `.

- [ ] ~~only calculate bin if the numeric field is not set to `ordinal` (See #88)~~

- [x] For `bin`, enumerate a fix set of `maxbins`  (maybe `10`, `20`, `40`) and apply these different bin parameters to the bin 
  - [x] add a way to calculate binned field stats for a certain parameter 
  - [x] when we call stats, read if we have it cached, it not calculate and cache. 

- [x] For `timeUnit`, enumerate all possible `timeUnit`, starting from single-part timeUnit (month, day) and maybe skip some of the composite timeUnit if stats from single-part timeUnit suggests that we can skip.
  - [ ] ~~Store the default `timeUnit` in `schema` the largest timeUnit with multiple values default `timeUnit` (as suggested in https://github.com/vega/vega-lite-ui/issues/47)~~
  - [ ] ~~maybe need to store what are available `timeUnit` in the schema so application can know.~~

Blocked by #88

RileyChang:
Fix #95 
1. `cardinality` method functions correctly for bin/timeUnit fields (issue #95)
2. Test coverage for cardinality


RileyChang:
Currently, tests are failing due to a single `nest` test. I believe it fails because it depends on the `fixture` class, which has become a bit outdated. Might open separate issue for fixing fixture?

RileyChang:
@kanitw Currently, all `TimeUnit`s supported by the datalib date bin method are being enumerated in `Schema.build`. Special units like `yearmonthday` are simply left up to datalib to infer an appropriate unit. I can't remember how we decided to handle these special types, is how I'm doing it now sufficient?

kanitw:
Looks like this one is conflicting with the other PR I already merge.  Maybe you want to resolve conflicts first?

RileyChang:
Merged with master. Tests are currently failing, but that should be fixed with felix's PR #144, which should probably be merged first (just fixes a typo).

src/schema.ts
```diff
@@ -171,9 +186,39 @@ function binSummary(maxbins: number, summary: Summary) {
...
+  switch (unit) {
+    // these are the units that dl.bins.date supports
+    case TimeUnit.YEAR:
+    case TimeUnit.MONTH:
+    case TimeUnit.DAY:
+    case TimeUnit.HOURS:
+    case TimeUnit.MINUTES:
+    case TimeUnit.SECONDS:
+      // FIXME: call date function correctly? This syntax is odd.
+      bin = dlBin['date']({
```
kanitw: Better correctly add this method to the interface file

test/schema.test.ts
```diff
@@ -174,6 +174,31 @@ describe('schema', () => {
...
+        {a: '1/1/2016'},
+        {a: '1/1/2026'}
+      ];
+      const cardinalitySchema = Schema.build(cardinalityData);
+      const cardinality: number = cardinalitySchema.cardinality({
+        field: 'a',
+        channel: Channel.X,
+        timeUnit: 'yearmonthdate'
+      });
+      assert.equal(cardinality, 10);
```
kanitw: I dont understand why this should be 10.  Can u explain why?

RileyChang:
@kanitw For some weird reason I can't seem to import the `bins` or `date` method from datalib like I did with `inferAll` for instance. I tried defining it in the interface file the exact same way. The best I could do was use `require`, which doesn't seem like the way to go.

test/schema.test.ts
RileyChang: Datalib doesn't support 'yearmonthdate' as far as I know, so in this case it will infer the `year` unit. Since 2026 - 2016 = 10 and datalib decides a step size of 1, 10 should be the cardinality. Is that incorrect?

kanitw: Doesn't sound right to me, but maybe we can file an issue and fix this in a follow up issue?

kanitw:
> @kanitw For some weird reason I can't seem to import the bins or date method from datalib like I did with inferAll for instance. I tried defining it in the interface file the exact same way. The best I could do was use require, which doesn't seem like the way to go.

What's the error when you tried to do so?

RileyChang:
@kanitw The error is "'datalib/src/bins/bins has no exported member 'date'". This happens even when I declare the module in the typings file, just like the other ones.

src/schema.ts
```diff
@@ -163,17 +180,99 @@ export class Schema {
...
+  // have the timeUnit bin scheme, need to determine new stats
+  // start with summary, pre-binning
+  const result = extend({}, summary);
+
+  const newUnique = binUnique(bin, summary.unique);
+  const values = uniqueToValues(newUnique);
+
+  result.unique = newUnique;
+  // 'count', 'valid', and 'missing' don't change
+  result.distinct = (type === Type.ORDINAL) ? keys(result.unique).length : (bin.stop - bin.start) / bin.step;
```
kanitw: I think this is never ordinal?  -- you only call this for bin of Q, timeUnit of T.  So I guess we can remove `type` from all relevant methods.

src/schema.ts
```diff
@@ -163,17 +180,99 @@ export class Schema {
...
+  result.min = bin.start;
+  result.max = bin.stop;
+  result.mean = dl.mean(values);
```
kanitw: These statistics can be copied directly from the original data.  

Currently you calculate stats from only the unique values, which doesn't represent the true distribution of the data.

src/schema.ts
```diff
@@ -163,17 +180,99 @@ export class Schema {
...
+      });
+  }
+
+  return binStats(bin, summary, type);
+}
+
+/**
+ * @return a new summary based on a new binning scheme and old summary statistics
+ */
+function binStats(bin, summary: Summary, type: Type): Summary {
```
kanitw: Please add appropriate interface for `bin` ... right now it's quite hard to read.

RileyChang:
@kanitw I addressed your comments, but the interface I added for `Bin` is kind of weird because it's not very well documented in Datalib. Let me know if you see a better way to do it.

src/schema.ts
```diff
@@ -312,3 +288,17 @@ export interface FieldSchema {
...
+interface Bin {
+  start: number | Date;
+  stop: number | Date;
+  step: number;
+  value(value: number | Date): number | Date;
+  index(value: number | Date): number;
+  unit: Unit;
+}
+
+interface Unit {
```
kanitw: BinUnit?

src/schema.ts
```diff
@@ -312,3 +288,17 @@ export interface FieldSchema {
...
+  date?();
```
kanitw: Doesn't it return anything?

src/schema.ts
```diff
@@ -234,45 +234,34 @@ function timeUnitSummary(unit: TimeUnit, summary: Summary, type: Type) {
...
   // have the timeUnit bin scheme, need to determine new stats
   // start with summary, pre-binning
   const result = extend({}, summary);
 
-  const newUnique = binUnique(bin, summary.unique);
-  const values = uniqueToValues(newUnique);
-
-  result.unique = newUnique;
-  // 'count', 'valid', and 'missing' don't change
-  result.distinct = (type === Type.ORDINAL) ? keys(result.unique).length : (bin.stop - bin.start) / bin.step;
```
kanitw: Now I recall why you add type.  It's actually not `type`, but rather `scaleType`.  But I guess let's ignore those for now.

src/schema.ts
```diff
@@ -312,3 +288,17 @@ export interface FieldSchema {
   timeStats?: {[timeUnit: string]: Summary};
   title?: string;
 }
+
+interface Bin {
```
kanitw: This is what datalib returns?  If so, I think this should be in datalib.d.ts.   Also add JsDocs style /** */ comment that this is return from which method.

src/schema.ts
```diff
@@ -234,45 +234,34 @@ function timeUnitSummary(unit: TimeUnit, summary: Summary, type: Type) {
...
  * @return a new unique object based off of the old unique count and a binning scheme
  */
-function binUnique(bin, oldUnique) {
+function binUnique(bin: Bin, oldUnique) {
   const newUnique = {};
   for (var value in oldUnique) {
     // date types are tricky, need to convert into correct unit
-    let bucket = bin.unit.date ? bin.value(bin.unit.unit(new Date(value))) : bin.value(value);
-    if (!newUnique[bin.value(value)]) {
+    let bucket: any = bin.unit.date ? bin.value(bin.unit.unit(new Date(value))) : bin.value(Number(value));
```
kanitw: isn't it always number?  (Or I guess at most it can be a date?  If so, use union type rather than `any`).

src/schema.ts
```diff
@@ -182,17 +199,80 @@ export class Schema {
...
-  binSum.distinct = (bin.stop - bin.start) / bin.step;
-  return binSum;
+
+  return binStats(bin, summary);
+}
+
+function timeUnitSummary(unit: TimeUnit, summary: Summary, type: Type): Summary {
+  var bin: Bin;
+  switch (unit) {
+    // these are the units that dl.bins.date supports
```
kanitw: Actually, DL's `months` = VL's `yearmonth`  and same for a few many others. 

See https://github.com/vega/datalib/wiki/Statistics#dl_bins_date for more info.

RileyChang: @kanitw Are you suggesting that we should add `case TimeUnit.YEARMONTH` and so on? Because the `dl.bins.date` method doesn't work with that unit.

kanitw: Why do you think it doesn't work?  It's just called differently.  
(If it really doesn't works and is buggy, file a bug in datalib.)

> Because the dl.bins.date method doesn't work with that unit.

You will need to convert Vega-Lite's multi-unit (e.g., `YEARMONTH`) to DL's corresponding pluralized periodic units (e.g., `months`). 

Some VL time units don't have 1:1 correspondents in DL, so we can ignore those cases for now.

kanitw: Btw, actually for `YEARMONTHDAY` and similar units without 1:1 correspondents in DL, if you return the original summary, which contains the original cardinality (`distinct`), that would be at least more accurate that inferred binned (which would become just `year`).

kanitw: To convert YEARMONTH to months and same for others, could you help add `toDataLibTimeUnit` in VL's `timeunit.ts`?   This would might become helpful for other projects.  

- [ ] Also notice that `hours` = `hour` in DL,  `datehours` = `hours`, `day` = `weekdays`, et.c

kanitw: Actually I was wrong about 1:1 mapping, see https://github.com/vega/vega-lite/issues/1487 for a correct mapping.

src/schema.ts
```diff
@@ -182,17 +199,80 @@ export class Schema {
...
+    case TimeUnit.YEAR:
+    case TimeUnit.MONTH:
+    case TimeUnit.DAY:
+    case TimeUnit.HOURS:
+    case TimeUnit.MINUTES:
+    case TimeUnit.SECONDS:
+      bin = dl.bins.date({
```
kanitw: Actually, the whole idea of using `dl.bins.date` is totally incorrect.  

I was curious, so I read and test it a bit

From the docs:

>  this method first determines an appropriate time unit (years, months, days, etc) and then uses dl.bin to produce a binning scheme.

It still perform binning, which means that it might not directly count # of values in between and might also adjust the start / stop.  I suggest that you add a number of test cases manually and see if the output is correct.  

I tried a few in the console of the datalib homepage and it doesn't seem correct. 

> var b = dl.bins.date({
  min: Date.parse('2000-01-01'),
  max: Date.parse('2005-12-31'), 
  unit: 'month' // equivalent to VL's `yearmonth`
}); 

I would except to get cardinality 5*12 = 60. However

> (b.stop - b.start)/b.step = 13
> (b.stop - b.start) = 78

Thus, neither of this is correct.

src/schema.ts
```diff
@@ -1,9 +1,12 @@
 import {Type} from 'vega-lite/src/type';
 import {Channel} from 'vega-lite/src/channel';
 import {autoMaxBins} from 'vega-lite/src/bin';
+import {TimeUnit, TIMEUNITS} from 'vega-lite/src/timeunit';
 import {summary} from 'datalib/src/stats';
 import {inferAll} from 'datalib/src/import/type';
-import * as dlBin from 'datalib/src/bins/bins';
+
+declare function require(name: string);
+var dl = require('datalib');
```
kanitw: Can't you use `import `bins` from 'datalib'` ?

src/schema.ts
```diff
@@ -1,7 +1,7 @@
 import {Type} from 'vega-lite/src/type';
 import {Channel} from 'vega-lite/src/channel';
 import {autoMaxBins} from 'vega-lite/src/bin';
-import {TimeUnit, TIMEUNITS} from 'vega-lite/src/timeunit';
+import {TimeUnit, TIMEUNITS, convertDate} from 'vega-lite/src/timeunit';
```
kanitw: I don't see a PR for convertDate in Vega-Lite yet.  
Maybe name it `applyTimeUnit`? 

(Ideally I want it to be `vl.timeUnit.apply` but `apply` is a native JS method.  Any better verb that works?  Maybe just `convert` so we get `vl.timeUnit.convert`? )

src/schema.ts
```diff
@@ -43,15 +46,30 @@ export class Schema {
         }
       } else if (primitiveType === PrimitiveType.DATE) {
         type = Type.TEMPORAL;
+        // need to get correct min/max of date data
```
kanitw: > need to get correct min/max of date data

Why?  Is it because datalib doesn't infer this correctly?  I thought you say it's not a problem anymore.

src/schema.ts
```diff
@@ -43,15 +46,30 @@ export class Schema {
...
+          }
+          if (time > (summary.max as Date).getTime()) {
+            summary.max = new Date(time);
+          }
+        }
       } else {
         type = Type.NOMINAL;
       }
-      return {
+      var fieldSchema: FieldSchema = {
```
kanitw: Why declaring this variable?

1) It seems like you're not recalculating any binStats at all although calculating the default bin would be helpful? 
2) I just realize I was wrong about not recalculating cardinalities for time units at all.  
We should pre-calculate them to support #22.

typings/datalib.d.ts
```diff
@@ -39,6 +39,22 @@ interface Summary {
...
 }
 
+/** Describes a binning scheme. This is returned by the dl.bins and dl.bins.date methods. */
+interface Bin {
+  start: number;
+  stop: number;
+  step: number;
+  value(value: number | Date): number | Date;
+  index(value: number | Date): number;
+  unit: BinUnit;
```
kanitw: I think we no longer use this part?

src/schema.ts
```diff
@@ -75,27 +93,12 @@ export class Schema {
 
     let schema = new Schema(fieldSchemas);
 
-    // calculate preset bins
+    // calculate preset bins for quantitative data
```
kanitw: Why this for loop should be done after instantiating `new Schema`?

src/schema.ts
```diff
@@ -210,6 +257,6 @@ export interface FieldSchema {
   primitiveType: PrimitiveType;
   stats: Summary;
   binStats?: {[key: string]: Summary};
-  timeStats?: {[timeUnit: string]: Summary};
+  timeUnitCardinalities?: {[timeUnit: string]: number};
```
kanitw: Why not stats?  At least we can store the new unique object as well?

src/schema.ts
RileyChang: Datalib's `inferAll` method works correctly, but datalib's `summary` method doesn't get the type correct. It considers the length of the date string when it determines the stats, so we have to manually figure out the min/max date.

kanitw: > Datalib's summary method doesn't get the type correct. It considers the length of the date string when it determines the stats, so we have to manually figure out the min/max date.

This sounds like the comment you should write down.   Just saying 

>  // need to get correct min/max of date data

doesn't help much because readers don't know why you need to do it.

test/schema.test.ts
```diff
@@ -186,6 +175,170 @@ describe('schema', () => {
       });
       assert.equal(cardinality, 7);
     });
+
+    it('should correctly compute new binned cardinality when binned cardinality is less than non-binned cardinality', () => {
+      const cardinalityData = [
+        {a: 0}, {a: 1},                 // bin 0-1
+        {a: 2}, {a: 2}, {a: 3}, {a: 3}, // bin 2-3
+        {a: 6}, {a: 7}                  // bin 6-7
```
kanitw: If I understand correctly, bin 4-5 is empty.  If so add a line 

```
// bin 4-5 is empty
```

test/schema.test.ts
```diff
@@ -186,6 +175,170 @@ describe('schema', () => {
...
+        {a: 'January  1, 2000'}
+      ];
+      let cardinalitySchema = Schema.build(cardinalityData);
+      let cardinality: number = cardinalitySchema.cardinality({
+        field: 'a',
+        channel: Channel.X,
+        timeUnit: 'quartermonth'
+      });
+      assert.equal(cardinalitySchema.primitiveType('a'), PrimitiveType.DATE);
+      assert.equal(cardinality, 3);
```
kanitw: Why is this 3?            Jan = Q1,   May & June = Q3?

typings/datalib.d.ts
```diff
@@ -39,6 +39,12 @@ interface Summary {
   modeskew: number;
 }
 
+/** Part of the Bin interface */
+interface BinUnit {
```
kanitw: Do we use BinUnit anywhere?

test/schema.test.ts
RileyChang: I believe this is correct because there are three different months, and month should "override" quarter? So the cardinality is 3?

kanitw: Oh true, it's quarter month, not quarter. My bad.

PR #136, opened 2016-07-14
Domain return type should not always be string array

RileyChang:
1. Fixes the `domain` method so that the array returned is of an appropriate type.
2. Revised and increased test coverage accordingly.


RileyChang:
The date case is a little tricky:
1. Currently using `new Date()` to coerce to `Date` type, only supports some date formats.
2. Dates of different formats can map to the same date, so there is currently a very inefficient looping algorithm to remove duplicates.

RileyChang:
@kanitw This might be ready for merging, what do you think about the 2 issues I mentioned above? Concerning (2), maybe it would be okay to just allow duplicates in the domain?

src/schema.ts
```diff
@@ -92,10 +92,29 @@ export class Schema {
...
+      // return [min, max] for quantitative data
+      domain = [fieldSchema.stats.min, fieldSchema.stats.max];
+    } else if (fieldSchema.primitiveType === PrimitiveType.INTEGER ||
+        fieldSchema.primitiveType === PrimitiveType.NUMBER) {
+      // coerce non-quantitative numerical data into number type
+      domain = domain.map(x => +x);
+    } else if (fieldSchema.primitiveType === PrimitiveType.DATE) {
+      // coerce dates into date type
+      domain = domain.reduce(function(result: Date[], curr) {
+        var date: Date = new Date(curr);
```
kanitw: Just like QUANTITATIVE, we only want `[min, max]` for domain.  
So if we introduce `min`, `max` calculation, this can easily benefit for that :)

kanitw:
> only supports some date formats.

This is okay.  Having to deal with all date format seems beyond scope of our work for now.  

For 2), just replied that I think we only want `[min, max]` for date.  (We might want unique values for timeUnits that we want ordinal values -- but we can ignore that for now.)

RileyChang:
@kanitw Changed the `summary` interface definition so that `min` and `max` are of type `any` now instead of `number` to accommodate temporal max and min types

src/schema.ts
```diff
@@ -51,7 +54,29 @@ export class Schema {
...
+        for (let maxbins of opt.maxBinsList) {
+          fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
+        }
+      } else if (fieldSchema.type === Type.TEMPORAL) {
+        // need to get min/max of date data
+        const millis = [];
+        for (var i = 0; i < data.length; i++) {
+          millis.push(new Date(data[i][fieldSchema.field]).getTime());
+        }
+        millis.sort();
```
kanitw: Sorting takes O(n log n). You can get min max in O(n).   (Just check if value is less than min and more than max in the loop.

src/schema.ts
```diff
@@ -51,7 +54,29 @@ export class Schema {
...
+    // calculate preset bins
+    for (let fieldSchema of fieldSchemas) {
+      if (fieldSchema.type === Type.QUANTITATIVE) {
+        fieldSchema.binStats = {};
+        for (let maxbins of opt.maxBinsList) {
+          fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
+        }
+      } else if (fieldSchema.type === Type.TEMPORAL) {
+        // need to get min/max of date data
+        const millis = [];
```
kanitw: maybe rename `millis` to `time` ?

src/schema.ts
```diff
@@ -84,18 +109,42 @@ export class Schema {
...
+      if (typeof encQ.bin === 'boolean') {
+        // autoMaxBins defaults to 10 if channel is EnumSpec
+        bin = {
+          maxbins: autoMaxBins(encQ.channel as Channel)
+        };
+      } else {
+        bin = encQ.bin;
+      }
+      const fieldSchema = this.fieldSchemaIndex[encQ.field as string];
+      if (!fieldSchema.binStats[bin.maxbins as string]) {
```
kanitw: Please declare `const maxbins: any = bin.maxbins` instead of having `as` every single time you use them.

typings/datalib.d.ts
```diff
@@ -29,8 +29,8 @@ interface Summary {
   valid: number;
   missing: number;
   distinct: number;
-  min: number;
-  max: number;
+  min: any;
```
kanitw: is it really `any`  or rather `number | string` or something like that?

src/schema.ts
```diff
@@ -51,7 +54,33 @@ export class Schema {
...
+          fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
+        }
+      } else if (fieldSchema.type === Type.TEMPORAL) {
+        // need to get min/max of date data
+        fieldSchema.stats.min = new Date(data[0][fieldSchema.field]);
+        fieldSchema.stats.max = new Date(data[0][fieldSchema.field]);
+        for (var i = 0; i < data.length; i++) {
+          const time = new Date(data[i][fieldSchema.field]).getTime();
+          if (time < (fieldSchema.stats.min as Date).getTime()) {
+            fieldSchema.stats.min = new Date(time);
```
kanitw: I'm not sure which format we should store date in yet, but i'll let go through for now and we can revisit this later.

PR #140, opened 2016-07-18
Add scalePropertiesSupportedByScaleType constraint

felixcodes:
Add constraint that ensures that scale properties must be supported by correct scale type.


src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
   },{
+    name: 'scalePropertiesSupportedByScaleType',
+    description: 'Scale properties must be supported by correct scale type',
+    properties: [Property.SCALE, Property.SCALE_TYPE, Property.TYPE],
+    allowEnumSpecForProperties: true,
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+        const hasScaleType: boolean = contains(keys(scale), Property.SCALE_TYPE);
```
kanitw: This is a terribly expensive way to check if `scale` has `type` property.  (O(n)!)  

Plus, it doesn't help you distinguish enumSpec from specific values.

src/constraint/spec.ts
```diff
@@ -538,7 +538,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
   {
     name: 'scaleZeroMustMatchScaleType',
```
kanitw: Shouldn't this be a part of `scalePropertiesSupportedByScaleType`?  

Plus, I don't know why it is in `spec.ts`.  

I think you should should this.

src/property.ts
```diff
@@ -168,3 +169,110 @@ export function getNestedEncodingPropertyChildren(parent: Property) {
 export function isNestedEncodingProperty(prop: Property) {
   return prop in NESTED_ENCODING_INDEX;
 }
+
+const SUPPORTED_SCALE_PROPERTY =
+[
+  // {
```
kanitw: @felixcodes  You said this PR is ready.  Yet there are these // everywhere.

test/constraint/encoding.test.ts
```diff
@@ -444,6 +444,31 @@ describe('constraints/encoding', () => {
...
   });
 
+  describe('scalePropertiesSupportedByScaleType', () => {
+    let encQ: EncodingQuery = {
+      channel: Channel.X,
+      field: 'A',
+      type: Type.QUANTITATIVE,
+      scale: {type: undefined, zero: true}
+    };
+
```
kanitw: You haven't tested any other properties.

src/constraint/spec.ts
```diff
@@ -534,30 +534,6 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
       }
       return true;
     }
-  },
-  {
-    name: 'scaleZeroMustMatchScaleType',
```
kanitw: A correct commit message would be "remove `scaleZeroMustMatchScaleType` constraint and its test."

src/query/encoding.ts
```diff
@@ -34,6 +34,7 @@ export interface BinQuery extends EnumSpec<boolean> {
 
 export interface ScaleQuery extends EnumSpec<boolean> {
   // TODO: add other properties from vegalite/src/scale
+  bandSize?: number | EnumSpec<number> | ShortEnumSpec;
```
kanitw: Have you added all properties?  If so, remove the TODO as it's done!

src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+        const hasScaleType: boolean = (scale.type !== undefined);
+
+        if ((!hasScaleType && isEnumSpec(encQ.type)) || isEnumSpec(scale.type)) {
+          return true;
+        }
+
+        for (let scaleProp in scale) {
+          if (SUPPORTED_SCALE_PROPERTY_INDEX[scaleProp]) {
```
kanitw: is there any scale property that's not in the `SUPPORTED_SCALE_PROPERTY_INDEX`?    

I can think of one case (guess what I can thought of) ?

src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+
+            const sType = scaleType(scale.type, encQ.timeUnit, encQ.type);
```
kanitw: Why do you have to calculate `sType` repeatedly for every single scale property?

src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
   },{
+    name: 'scalePropertiesSupportedByScaleType',
+    description: 'Scale properties must be supported by correct scale type',
+    properties: [Property.SCALE, Property.SCALE_TYPE, Property.TYPE],
+    allowEnumSpecForProperties: true,
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+        const hasScaleType: boolean = (scale.type !== undefined);
```
kanitw: Here `allowEnumSpecForProperties` is `true`.   

If you add a test where `encQ.scale = false` I think you will get error that can't get `type` of undefined.

src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+
+        if ((!hasScaleType && isEnumSpec(encQ.type)) || isEnumSpec(scale.type)) {
```
kanitw: Add comment explaining why each of these two cases we can ignore.

test/constraint/encoding.test.ts
```diff
@@ -444,6 +444,201 @@ describe('constraints/encoding', () => {
...
+  describe('scalePropertiesSupportedByScaleType', () => {
+    describe('scaleBandSize', () => {
+      let encQ: EncodingQuery = {
+        channel: Channel.X,
+        field: 'A',
+        type: Type.QUANTITATIVE,
+        scale: {bandSize: 10, type: undefined}
+      };
+
+      it('should return false if scaleTYpe does not suport scaleBandSize', () => {
```
kanitw: scaleTYpe ??? ... clean up

test/constraint/encoding.test.ts
```diff
@@ -445,27 +445,197 @@ describe('constraints/encoding', () => {
...
+        field: 'A',
+        type: Type.QUANTITATIVE,
+        scale: {bandSize: 10, type: undefined}
+      };
 
-    it('should return false if scaleType does not support scaleZero', () => {
-      [ScaleType.LOG, ScaleType.ORDINAL, ScaleType.TIME, ScaleType.UTC].forEach((scaleType) => {
-        (encQ.scale as ScaleQuery).type = scaleType;
-        assert.isFalse(ENCODING_CONSTRAINT_INDEX['scalePropertiesSupportedByScaleType'].satisfy(encQ, schema, defaultOpt));
+      it('should return false if scaleTYpe does not suport scaleBandSize', () => {
```
kanitw: You never test for a case where scale type is `undefined` and we infer type from scale type.   
(You don't have to add this for all properties, but at least should test it once.) 

Please utilize the test coverage tool to see if you coverage all **branches** (not just statement) for `scalePropertiesSupportedByScaleType` in `src/constraint/encoding.ts`

src/constraint/encoding.ts
```diff
@@ -140,6 +140,34 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
       return true;
     }
   },{
+    name: 'scalePropertiesSupportedByScaleType',
+    description: 'Scale properties must be supported by correct scale type',
+    properties: [Property.SCALE, Property.SCALE_TYPE, Property.TYPE],
```
kanitw: This actually is affected by all scale's child properties.  
Don't type them manually, leverage `NESTED_ENCODING_PROPERTY_PARENT_INDEX`.

kanitw: Please also add a few generate tests to ensure this works correctly.  (A part of the reason you miss this bug is that you don't do generate test.)

src/constraint/encoding.ts
```diff
@@ -141,6 +141,39 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+    description: 'Scale properties must be supported by correct scale type',
+    properties: SCALE_PROPERTIES.concat([Property.SCALE, Property.TYPE]),
+    allowEnumSpecForProperties: true,
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+        const hasScaleType: boolean = (scale.type !== undefined);
+
+        /**
```
kanitw: Incorrect type of comment

`/** */` is JSDoc -- for describing class / variable or method.

In this case, you wanna use  `//` or `/* */` (We generally use one or more lines of the former.)

src/constraint/encoding.ts
```diff
@@ -141,6 +141,39 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
   },{
+    name: 'scalePropertiesSupportedByScaleType',
+    description: 'Scale properties must be supported by correct scale type',
+    properties: SCALE_PROPERTIES.concat([Property.SCALE, Property.TYPE]),
+    allowEnumSpecForProperties: true,
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+        const hasScaleType: boolean = (scale.type !== undefined);
```
kanitw: hasScaleType is only used once.  I don't think we need this declaration line?

test/constraint/encoding.test.ts
```diff
@@ -444,6 +444,201 @@ describe('constraints/encoding', () => {
...
+        assert.isTrue(ENCODING_CONSTRAINT_INDEX['scalePropertiesSupportedByScaleType'].satisfy(encQ, schema, defaultOpt));
+      });
+    });
+
+    describe('scaleClamp', () => {
+      let encQ: EncodingQuery = {
+        channel: Channel.X,
+        field: 'A',
+        type: Type.QUANTITATIVE,
+        scale: {clamp: true, type: undefined}
```
kanitw: You're setting type afterward anyway, I'm not sure why you set `type=undefined` here?   (Same for others.)

test/constraint/encoding.test.ts
```diff
@@ -444,6 +444,201 @@ describe('constraints/encoding', () => {
     });
   });
 
+  describe('scalePropertiesSupportedByScaleType', () => {
+    describe('scaleBandSize', () => {
+      let encQ: EncodingQuery = {
+        channel: Channel.X,
+        field: 'A',
+        type: Type.QUANTITATIVE,
```
kanitw: In this test suite for `scalePropertiesSupportedByScaleType`, it seems like type is always `Q`.  Are you confident that the test are correct for non Q type too?  If not, test it for at least one of the scale properties?

src/constraint/encoding.ts
```diff
@@ -141,6 +141,36 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+    strict: true,
+    satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
+      if (encQ.scale) {
+        const scale: ScaleQuery = encQ.scale as ScaleQuery;
+
+         //  If encQ.type is an EnumSpec and scale.type is undefined, it is equivalent
+         //  to scale type is EnumSpec. If scale type is an EnumSpec, we do not yet know
+         //  what the scale type is, and thus can ignore the constraint.
+
+        if ((!(scale.type !== undefined) && isEnumSpec(encQ.type)) || isEnumSpec(scale.type)) {
```
kanitw: `!(scale.type !== undefined)`  

Why do you have to make this so complicated?

PR #142, opened 2016-07-21
 Fix #96 - Test to make sure that each property's default name are unique

### LINKED ISSUE #96: Add a test to make sure that property's default name are unique
This will be a test for `getDefaultName` in `model.ts`

felixcodes:
Fix #96 : test for getDefaultName in model.ts

### DIFF
--- a/test/model.test.ts
+++ b/test/model.test.ts
@@ -6,8 +6,8 @@ import {Mark} from 'vega-lite/src/mark';
 import {Type} from 'vega-lite/src/type';
 
 import {DEFAULT_QUERY_CONFIG} from '../src/config';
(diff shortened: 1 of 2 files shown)


test/model.test.ts
```diff
@@ -21,6 +21,23 @@ describe('SpecQueryModel', () => {
     return SpecQueryModel.build(specQ, schema, DEFAULT_QUERY_CONFIG);
   }
 
+  describe('getDefaultName', () => {
+    it('should have no duplicate default names.', () => {
+      let props: Property[] = DEFAULT_PROPERTY_PRECENCE;
+
+      let unique = props.map((prop) => {
```
kanitw: Why do you need to count?   Why don't just assert that each key added should not already exist?

PR #144, opened 2016-07-21
fix typo precence --> precedence in model.test.ts

felixcodes:
@RileyChang #135


kanitw:
Why was this passing the test / lint before?

felixcodes:
I am not sure why this was passing lint. I definitely ran watch when I changed the other typos and it passed. Maybe I made a different change before I pushed and only looked at watch's npm run test result and not lint. Whatever the case -- my fault.

PR #141, opened 2016-07-19
Split query.ts

felixcodes:
Fix #121. Travis build will fail due to an unreleased dependency in VL. Passes all tests locally.

PR #148, opened 2016-07-25
Fix #86 Specify Schema in json

### LINKED ISSUE #86: Allow supplying schema and statistics as a part of the json
Sometimes default type inference (from datalib) might be incorrect.  
- [x] We should add `schema` property to `Query` as an object that takes key-value mapping (from `field: string` to a `FieldSchema` object)  but the `object` here might not contains all properties of `FieldSchema` for example users might want to override `type` of a field `x`, but keep the same `stats` and thus only provide `{x: {type: 'ordinal'}}`.
- [x] In `query.ts` check if the query has `schema`, if so duplicate the schema and override some default inferred schema properties with ones provided in `schema`.  
  - [ ] We might want to introduce `schema.duplicate` method for re-creating `schema` object to prevent side-effect. 

Hint: use the `extends` method in `util.ts`

RileyChang:
Allows a user to override Schema defaults by specifying schema properties in the query object.


RileyChang:
TODO: update `Schema.fieldSchemas` in the setter method as well. Maybe we can just get rid of the list of schemas and just keep the index?

src/query.ts
```diff
@@ -16,10 +16,21 @@ import {nest} from './nest';
...
 
 export function query(q: Query, schema: Schema, config?: Config) {
+  // 0. Override schema object with the properties specified by the query
+  if (q.schema) {
+    keys(q.schema).forEach(field => {
+      const fieldSchema: FieldSchema = schema.getSchema(field);
+      keys(q.schema[field]).forEach(prop => {
+        fieldSchema[prop] = q.schema[field][prop];
+      });
+      schema.setSchema(field, fieldSchema);
```
kanitw: I think you dont need this line since the fieldSchema object you get is mutable anyway?

RileyChang: The `getSchema()` method currently returns a duplicate of the `schema`, so I think I need to set it again. Perhaps I could change `getSchema()` so that it returns the actual `schema`?

kanitw:
@RileyChang  You'll have to migrate changes to query.ts and it's test file to the new location due to #141. Feel free to consult @felixcodes if it's unclear.

test/query/query.test.ts
```diff
@@ -65,6 +66,29 @@ describe('query/query', () => {
...
+      const q: Query = {
+        spec: {
+          data: {
+            values: [
+              {x: '1/1/2000'}
+            ]
+          },
+          mark: 'point',
+          encodings: [
+            {channel: Channel.X, field: 'x', type: Type.TEMPORAL}
```
kanitw: change type here to enumSpec and check that the only enumerated spec is nominal.

kanitw:
THis won't work with the new table schema code anymore, so I'm closing.

PR #154, opened 2016-07-29
Add support for grouping by array of property types. 	

### LINKED ISSUE #122: More generalize and more understandable `groupBy` method
Current `groupBy` syntax in a CompassQL query looks like this: 

```js
{
  "spec": ...,
  "nest": [
    {"groupBy": "field"},
    {"groupBy": "fieldTransform"},
    {"groupBy": "encoding", "orderGroupBy": "effectiveness"},
    {"groupBy": "markAndTranspose", "orderGroupBy": "effectiveness"}
  ],
  "orderBy/chooseBy": "effectiveness",
  ...
}
```

The problem is that naming each `groupBy` key function to be information  can be quite confusing (e.g., `encoding`, `markAndTranspose` are pretty bad name). 

Instead of providing arbitrary name like above, I propose that we have `groupBy` as an array of property names with possibly transformation for further grouping (like how I group channel below) instead

```js
{
  "spec": ...,
  "nest": [
    /** replacing {"groupBy": "field"} */
    {"groupBy": [{property: "field"}]},

    /** replacing {"groupBy": "fieldTransform"} */
    {"groupBy": ["aggregate", "timeUnit", "bin"]},  
    // - "field" will be automatically included at this level  since we include it in the parent level

    /** replacing {"groupBy": "encoding", "orderGroupBy": "effectiveness"} */
    {
      "groupBy": [
        {
          property: "channel",  
          keys: {
             xy: ["x", "y"], 
             facet: ["row", "column"], 
             style: ["color", "size", "shape", "opacity"] 
          }
        }
      ],   
      // -  "field", "aggregate", "timeUnit", "bin" will be automatically included at this level since 
      //    we include them in the parent levels
      // - ":position, facet, style" suffix after "channel" means that 
      //   instead of using raw channel, we transform channel like
      //   x/y => "position", row/column => "facet", and  color/size/shape/opacity => "style"
      //   before using the transformed channel to group things.
      //   Therefore, x and y will be considered equivalent because 
      //   the "channel" value will become "position" instead of the original "x" or "y". 
      //   Similar for row/column => "facet" and color/size/shape/opacity => "style".
      "orderGroupBy": "effectiveness"
    },

    /** replacing {"groupBy": "markAndTranspose", "orderGroupBy": "effectiveness"} **/
    {
      "groupBy": [
         "mark",
         {property: "channel",  keys: { xy: ["x", "y"],  facet: ["row", "column"] }}
     ], 
      // - "field", "aggregate", "timeUnit", "bin" will be automatically included at this level 
      //   since we include them the parent levels  
      // - We specify channel with different transform, so it's overridden.
      //   Basically, channel transforms here doesn't include "style" 
      //   so "color/size/shape/opacity" are not grouped together at this level
    },
  ],
  "orderBy/chooseBy": "effectiveness",
  ...
}
```

This way, our groupBy framework can work with new properties (e.g., `scaleType`) that we are adding as well.  

@jheer, @domoritz, @ZeningQu  any opinions?

kanitw:
Fix #122. 

See each commit for details.


kanitw:
Group name actually doesn't show up correctly in vega-editor yet.

vega_editor

@felixcodes  Do you want to help investigate why?

PR #157, opened 2016-07-29
Fix #155 Order the FieldSchemas

### LINKED ISSUE #155: Order Enumerated Fields
Currently enumerated fields use original order in the data.  We should allow the schema to be ordered by type then by name to produce more predictable enumeration order. 

vega_editor

(This is a critical issue, please try to solve this before others -- we can chat about it in detail later today.)

RileyChang:
Orders FieldSchemas in the following order:
1. By type. nominal < temporal < quantitative < ordinal
2. By field name, alphabetically

This should mean that when fields are enumerated, they should be presented in a more intuitive manner (issue #155).

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -54,6 +54,20 @@ export class Schema {
       };
     });
 
+    // order the fieldSchemas (sort them)
+    const order = [Type.NOMINAL, Type.TEMPORAL, Type.QUANTITATIVE, Type.ORDINAL];
+    fieldSchemas.sort(function(a: FieldSchema, b: FieldSchema) {
+      // first order by type: nominal < temporal < quantitative < ordinal
+      if (order.indexOf(a.type) < order.indexOf(b.type)) {
+        return -1;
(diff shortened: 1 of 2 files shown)


src/schema.ts
```diff
@@ -54,6 +54,20 @@ export class Schema {
       };
     });
 
+    // order the fieldSchemas (sort them)
+    const order = [Type.NOMINAL, Type.TEMPORAL, Type.QUANTITATIVE, Type.ORDINAL];
```
kanitw: N < O < T < Q

src/schema.ts
```diff
@@ -54,6 +54,20 @@ export class Schema {
...
+    fieldSchemas.sort(function(a: FieldSchema, b: FieldSchema) {
+      // first order by type: nominal < temporal < quantitative < ordinal
+      if (order.indexOf(a.type) < order.indexOf(b.type)) {
```
kanitw: Calling IndexOf Takes O(N) -- use dictionary would be more efficient

PR #165, opened 2016-08-02
 Convert EnumSpecIndex into a class

### LINKED ISSUE #164: Convert enumSpecIndex into a class.
It's always generated at run time, so we can leverage class to provide class methods to

- [x] Add `isEmpty()` and add tests that enumerate all properties and leverage `setEnumSpecIndex` to test 
- [x] `setEnumSpecIndex` => `set`
- [x] `hasPropertyIndex` => `hasProperty`
- [x] getters for `mark` and `encodings` and `encodingIndicesByProperty`

felixcodes:
Fix #164 - Refactor EnumSpecIndex to be a class with relevant tests.


src/enumspecindex.ts
```diff
@@ -0,0 +1,134 @@
...
+  // TODO: transform
+
+  /**
+   * Dictionary mapping encoding index to an encoding enum spec index.
+   */
+
+  private _encodings: EncodingsEnumSpecIndex;
+  private _encodingIndicesByProperty: Dict<number[]>;
+
+  constructor(mark: EnumSpec<Mark>, encodings: EncodingsEnumSpecIndex, encodingIndicesByProperty: Dict<number[]>) {
```
kanitw: What's the point of adding these parameters, if the only case is to set them to undefined and {}.  

Remove the params and just initialize them correctly.

src/enumspecindex.ts
```diff
@@ -0,0 +1,134 @@
...
+    this._mark = mark;
+    this._encodings = encodings;
+    this._encodingIndicesByProperty = encodingIndicesByProperty;
+  }
+
+  public set(index: number, prop: Property, enumSpec: EnumSpec<any>) {
```
kanitw: rename `set` to `setEncodingProperty` (more precise)

src/enumspecindex.ts
kanitw: Ok.  Now I see that you use them for the test.  Still, you should make these params optional with default values.

kanitw: I think let's remove these parameter anyway.  Using constructor to cheat might produce an enumSpecIndex that's not practical in practice.

test/enumspecindex.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  describe('hasProperty', () => {
+    it('should return true if encodingIndicesByProperty contains a specified encoding property', () => {
+      const enumSpecIndex = new EnumSpecIndex(undefined, {}, {'scale': [1,2]});
+      assert.equal(enumSpecIndex.hasProperty(Property.SCALE), true);
+    });
+
+    it('should return false if encodingIndicesByProperty does not contain a specified encoding property', () => {
+      const enumSpecIndex = new EnumSpecIndex(undefined, {}, {'channel': [1,2]});
```
kanitw: This enumSpec never occurs in real life. How can the `encodingIndicesByProperty` has `channel` but `encodings` has nothing?   So better generate things you wanna test using `setEncodingProperty` and mark's setter.

src/enumspecindex.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    /* istanbul ignore next */
+    throw new Error('Unimplemented for property ' + prop);
+  }
+
+  public isEmpty() {
+    return !this.mark && keys(this.encodingIndicesByProperty).length === 0;
+  }
+
+  public set mark(mark) {
```
kanitw: Since we'll have `setEncodingProperty`,  maybe let's make this a normal method (`setMark()`) for consistency.  

Also make both of these methods support chaining just like D3 (return `this`) and use it for testing.

test/enumspecindex.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import {SHORT_ENUM_SPEC} from '../src/enumspec';
+import {Property} from '../src/property';
+
+import {assert} from 'chai';
+
+describe('enumspecindex', () => {
+  describe('isEmpty', () => {
+    it('should return false if encoding property is set', () => {
+      let enumSpecIndex = new EnumSpecIndex()
+        .setEncodingProperty(0, Property.SCALE, SHORT_ENUM_SPEC);
```
kanitw: I don't know why TS doesn't complain yet, but we never set these to SHORT_ENUM_SPEC and it doesn't match the type.  Better fix them with full enum spec; otherwise, it might throw error when we upgrade to future version of TS.

PR #177, opened 2016-08-04
Add basic support for encoding sort

### LINKED ISSUE #150: Basic Support for Encoding.sort
- [x] Basic Support
- [x] ~~Enumeration~~ Forked as #172

felixcodes:
fix #150 
  - Add sort property and sort child properties
  - Support encoding.sort in SpecModel.toSpec()
  - Support shorthand with sort

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -40,6 +40,12 @@ export interface QueryConfig {
   // TODO: Come back and implement correctly when designing sort enumeration.
   sorts?: SortOrder[];
 
+  sortFields?: string[];
+
+  sortOps?: AggregateOp[];
+
+  sortOrders?: SortOrder[];
+
   scaleBandSizes?: number[];
 
   scaleDomains?: Array<number[] | string[]>;
@@ -110,6 +116,9 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   // TODO: Come back and implement correctly when designing sort enumeration.
   sorts: [SortOrder.ASCENDING, SortOrder.DESCENDING],
+  sortFields: [undefined],
+  sortOps: [undefined, AggregateOp.MEAN],
+  sortOrders: [SortOrder.ASCENDING, SortOrder.DESCENDING],
 
   scaleBandSizes: [17, 21],
   scaleDomains: [undefined],
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,7 @@ import {ScaleType} from 'vega-lite/src/scale';
 import {TimeUnit} from 'vega-lite/src/timeunit';
 import {Type} from 'vega-lite/src/type';
 
+import {SortOrder} from './query/encoding';
 import {Property, DEFAULT_PROPERTY_PRECEDENCE} from './property';
 
 export interface QueryConfig {
@@ -35,6 +36,9 @@ export interface QueryConfig {
 
   /** Default maxbins to enumerate */
   maxBinsList?: number[];
+
+  // TODO: Come back and implement correctly when designing sort enumeration.
+  sorts?: SortOrder[];
 
   scaleBandSizes?: number[];
 
@@ -103,6 +107,9 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   types: [Type.NOMINAL, Type.ORDINAL, Type.QUANTITATIVE, Type.TEMPORAL],
 
   maxBinsList: [5, 10, 20],
+
+  // TODO: Come back and implement correctly when designing sort enumeration.
+  sorts: [SortOrder.ASCENDING, SortOrder.DESCENDING],
 
   scaleBandSizes: [17, 21],
   scaleDomains: [undefined],
--- a/src/config.ts
+++ b/src/config.ts
@@ -116,7 +116,6 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   // TODO: Come back and implement correctly when designing sort enumeration.
   sorts: [SortOrder.ASCENDING, SortOrder.DESCENDING],
-  sortFields: [undefined],
   sortOps: [undefined, AggregateOp.MEAN],
   sortOrders: [SortOrder.ASCENDING, SortOrder.DESCENDING],
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -3,10 +3,10 @@ import {AggregateOp} from 'vega-lite/src/aggregate';
 import {Config} from 'vega-lite/src/config';
 import {Mark} from 'vega-lite/src/mark';
 import {ScaleType} from 'vega-lite/src/scale';
+import {SortOrder} from 'vega-lite/src/sort';
 import {TimeUnit} from 'vega-lite/src/timeunit';
 import {Type} from 'vega-lite/src/type';
 
-import {SortOrder} from './query/encoding';
 import {Property, DEFAULT_PROPERTY_PRECEDENCE} from './property';
 
 export interface QueryConfig {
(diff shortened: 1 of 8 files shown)


kanitw:
I don't see anything related to `fromSpec`, at least you have to add/modify the test to consider a case with different type of sort.

src/config.ts
```diff
@@ -104,6 +114,12 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxBinsList: [5, 10, 20],
 
+  // TODO: Come back and implement correctly when designing sort enumeration.
+  sorts: [SortOrder.ASCENDING, SortOrder.DESCENDING],
+  sortFields: [undefined],
```
kanitw: This should be similar to fields, which is basically drawn from the schema.

src/query/shorthand.ts
```diff
@@ -121,6 +121,13 @@ export function fieldDef(encQ: EncodingQuery,
         value: value(encQ.bin['maxbins'], replace[Property.BIN_MAXBINS])
       });
     }
+  } else if (include[Property.SORT] && encQ.sort && !isEnumSpec(encQ.sort)) {
+    if (!(encQ.sort as SortFieldQuery)['op']) {
```
kanitw: why not `.op`?

kanitw: I'm also not sure why we need this branch for sort but not for scale?  What's different?  (I mean you might be right, but can you explain?)

src/query/encoding.ts
```diff
@@ -21,13 +21,34 @@ export interface EncodingQuery {
   bin?: boolean | BinQuery | ShortEnumSpec;
   scale?: boolean | ScaleQuery | ShortEnumSpec;
 
+  sort?: SortOrder | SortFieldQuery | ShortEnumSpec;
```
kanitw: The tricky part is here.  

We have to decide whether it's 

`SortOrder | Enumspec<SortOrder> | SortFieldQuery | ShortEnumSpec`

or 

`SortOrder | SortQuery | ShortEnumSpec` 

where SortQuery supports enumerating both SortOrder and SortField. 

The latter might be more correct, but also more complicated to handle.

kanitw: We don't really have to handle this question right now.  But if you don't want to do that,  don't introduce `SortFieldQuery` yet.  Simply use `SortOrder | SortField`like Vega-Lite because that obviously means we don't care about enumerating it yet.  

Introducing SortFieldQuery without caring about how to structure enumeration seems weird.

src/query/shorthand.ts
kanitw: To keep the structure consistent, the right location for this case is below 
```
else if (encQ[nestedPropParent] === false || encQ[nestedPropParent] === null) {
```
case

There you can check if `encQ[nestedPropParent]` is string.   (And add comment that for `sort`, it can be a string.)

This way, `sort` is always handled as a `nestedPropParent` rather than having two different code parts.

src/query/shorthand.ts
```diff
@@ -141,7 +134,12 @@ export function fieldDef(encQ: EncodingQuery,
   // TODO: axis, legend
   for (const nestedPropParent of [Property.SCALE, Property.SORT]) {
     if (include[nestedPropParent]) {
-      if (encQ[nestedPropParent] && !isEnumSpec(encQ[nestedPropParent])) {
+      if (typeof encQ[nestedPropParent] === 'string') {
```
kanitw: 1) Using isString from datalib for consistency
2) Put it below this clause.  You wanna split between when these properties really have nested values and when it's not rather than having the nested case in between.

src/query/encoding.ts
```diff
@@ -21,13 +21,34 @@ export interface EncodingQuery {
...
+  sort?: SortOrder | SortField;
+
   field?: Field | EnumSpec<Field> | ShortEnumSpec;
   type?: Type | EnumSpec<Type> | ShortEnumSpec;
   // TODO: value
 
   // TODO: axisQuery, legendQuery
 }
 
+export enum SortOrder {
```
kanitw: These are not different from Vega-Lite at all, why don't you just import from Vega-Lite? 
(My previous comment already suggests this.)

src/config.ts
```diff
@@ -104,6 +114,11 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxBinsList: [5, 10, 20],
 
+  // TODO: Come back and implement correctly when designing sort enumeration.
+  sorts: [SortOrder.ASCENDING, SortOrder.DESCENDING],
+  sortOps: [undefined, AggregateOp.MEAN],
```
kanitw: Op isn't optional (yet!).  So it can't be `undefined`.   Do MIN and MEAN

src/config.ts
```diff
@@ -104,6 +114,11 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxBinsList: [5, 10, 20],
 
+  // TODO: Come back and implement correctly when designing sort enumeration.
```
kanitw: I don't see you add a TODO in the follow up issue yet.  Very likely you will forget it if you don't.

test/query/spec.test.ts
```diff
@@ -232,6 +233,53 @@ describe('query/spec', () => {
...
     });
 
+    it('should produce correct SpecQuery with sort: SortOrder', () => {
+      const specQ = fromSpec({
+        data: {values: [{x: 1}, {x: 2}]},
+        transform: {filter: 'datum.x ===2'},
+        mark: Mark.POINT,
+        encoding: {
+          x: {field: 'x', sort: SortOrder.ASCENDING, type: Type.QUANTITATIVE},
+          y: {field: 'x', type: Type.QUANTITATIVE, scale: null}
```
kanitw: Y isn't relevant to this test.  You can either combine two test cases and use different type of sort in x and y or just put one channel.  

I prefer the former.

src/query/shorthand.ts
```diff
@@ -134,12 +135,14 @@ export function fieldDef(encQ: EncodingQuery,
...
   for (const nestedPropParent of [Property.SCALE, Property.SORT]) {
     if (include[nestedPropParent]) {
-      if (typeof encQ[nestedPropParent] === 'string') {
-        props.push({
-          key: nestedPropParent + '',
-          value: encQ[nestedPropParent]
-        });
-      } else if (encQ[nestedPropParent] && !isEnumSpec(encQ[nestedPropParent])) {
+      if (encQ[nestedPropParent] && !isEnumSpec(encQ[nestedPropParent])) {
+        if (isString(encQ[nestedPropParent])) {
```
kanitw: Missing comment explaining what is this `isString` case?   (Can you read my comment carefully?  Again, it's not fun to keep repeating similar comments.)

test/query/shorthand.test.ts
```diff
@@ -4,8 +4,8 @@ import {Mark} from 'vega-lite/src/mark';
 import {ScaleType} from 'vega-lite/src/scale';
 import {TimeUnit} from 'vega-lite/src/timeunit';
 import {Type} from 'vega-lite/src/type';
+import {SortOrder} from 'vega-lite/src/sort';
```
kanitw: Import should be order alphabetically.  (vl/src/sort should be above timeUnit)

src/query/encoding.ts
```diff
@@ -3,6 +3,7 @@ import {AggregateOp} from 'vega-lite/src/aggregate';
 import {defaultScaleType, TimeUnit} from 'vega-lite/src/timeunit';
 import {Type} from 'vega-lite/src/type';
```
kanitw: reorder imports?

PR #182, opened 2016-08-05
omitTableWithOcclusionIfAutoAddCount constraint

### LINKED ISSUE #179: Change `omitTableWithOcclusion` => `omitTableWithOcclusionIfAddAutoCount`
Without `autoCount: true`, it seems like this is not always the best idea.

felixcodes:
Fix #179 
- Add autoAddCount condition to constraint
- Update constraint name to reflect autoAddCount condition
- Add query test that interacts with omitTableWithOcclusionIfAutoAddCount constraint

### DIFF
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -508,7 +508,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     strict: false,
     satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
       if(opt.autoAddCount) {
-        // TODO: add mark properties channel into account
+        // TODO(#186): take mark properties channel into account
         if (specM.isDimension(Channel.X) &&
           specM.isDimension(Channel.Y) &&
           !specM.isAggregate() // TODO: refactor based on statistics
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -502,7 +502,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
   },
   {
     name: 'omitTableWithOcclusionIfAutoAddCount',
-    description: 'Plots without aggregation or autocount where x and y are both dimensions should be omitted',
+    description: 'Plots without aggregation or autocount where x and y are both dimensions should be omitted if autoAddCount is enabled',
     properties: [Property.CHANNEL, Property.TYPE, Property.TIMEUNIT, Property.BIN, Property.AGGREGATE, Property.AUTOCOUNT],
     allowEnumSpecForProperties: false,
     strict: false,
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -502,12 +502,13 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
   },
   {
     name: 'omitTableWithOcclusionIfAutoAddCount',
-    description: 'Plots without aggregation or autocount where x and y are both dimensions should be omitted if autoAddCount is enabled',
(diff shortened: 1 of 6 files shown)


src/constraint/spec.ts
```diff
@@ -501,17 +501,19 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
-    name: 'omitTableWithOcclusion',
-    description: 'Raw Plots with x and y are both dimensions should be omitted as they often lead to occlusion.',
+    name: 'omitTableWithOcclusionIfAutoAddCount',
+    description: 'Plots without aggregation or autocount where x and y are both dimensions should be omitted if autoAddCount is enabled',
```
kanitw: I don't understand why you have to cut "as they often lead to occlusion"?

src/constraint/spec.ts
```diff
@@ -501,17 +501,19 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
     properties: [Property.CHANNEL, Property.TYPE, Property.TIMEUNIT, Property.BIN, Property.AGGREGATE, Property.AUTOCOUNT],
     allowEnumSpecForProperties: false,
     strict: false,
     satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
-      if (specM.isDimension(Channel.X) &&
+      if(opt.autoAddCount) {
+        if (specM.isDimension(Channel.X) &&
```
kanitw: add // TODO: taking mark properties channel into account (Just for noting)

test/query/query.test.ts
```diff
@@ -65,6 +65,29 @@ describe('query/query', () => {
       assert.equal((<SpecQueryModel>result.items[0]).specQuery.mark, 'tick');
       assert.equal((<SpecQueryModel>result.items[1]).specQuery.mark, 'point');
     });
+
+    it('should return a plot with name on x', () => {
```
kanitw: So many problems with this line!
1) The test description is too low level, doesn't reflect the test goal. 
2) Why do you put it under "nested query" describe, which is not relevant to this test at all.

test/generate.test.ts
```diff
@@ -142,7 +142,7 @@ describe('generate', function () {
         }]
       };
 
-      const answerSet = generate(query, schema);
+      const answerSet = generate(query, schema, CONFIG_WITH_AUTO_ADD_COUNT);
```
kanitw: This line seems random.

kanitw: Can you explain why?

test/constraint/spec.test.ts
```diff
@@ -952,8 +952,8 @@ describe('constraints/spec', () => {
     });
   });
 
-  describe('omitTableWithOcclusion', () => {
-    it('return false for raw plot with both x and y as dimensions.', () => {
+  describe('omitTableWithOcclusionIfAutoAddCount', () => {
+    it('return false for raw plot with both x and y as dimensions without autocount or aggregation.', () => {
```
kanitw: Instead of adding random testing in `query.ts`, why don't you add a proper test here and in `generate.test.ts`?

test/generate.test.ts
felixcodes: slacked explanation (and fixed the missing description updated)

kanitw: "Explained in Slack"..

test/generate.test.ts
```diff
@@ -155,6 +155,27 @@ describe('generate', function () {
     });
   });
 
+  describe('NxO', () => {
```
kanitw: why isn't this in 2D?

PR #183, opened 2016-08-05
Fix ranking to make spec return [ ] if answerSet is empty

### LINKED ISSUE #178: Why does this spec returns only `[undefined]`?
```
{
  "spec": {
    "mark": "?",
    "encodings": [
      {
        "channel": "y",
        "field": "Cylinders",
        "type": "ordinal"
      },
      {
        "field": "Name",
        "type": "nominal",
        "channel": "?"
      }
    ],
    "config": {
      "overlay": {"line": true},
      "scale": {"useRawDomain": true},
      "facet": {"cell": {"width": 150,"height": 150}}
    }
  },
  "chooseBy": "effectiveness",
  "config": {
    "verbose": true
  }
}
```

If nothing qualifies, it should return `[]`.   

Note: from quick inspection.  `answerSet` is `[]`, so the bug is probably in either nesting or ranking.

felixcodes:
Fix #178 
- Fix ranking to make spec return [ ] if answerSet is empty
- test rank method with empty input group

### DIFF
--- a/test/ranking/ranking.test.ts
+++ b/test/ranking/ranking.test.ts
@@ -23,7 +23,7 @@ describe('ranking', () => {
               {channel: Channel.SHAPE, field: 'N', type: Type.NOMINAL},
             ]
           },
-          chooseBy: "effectiveness"
+          chooseBy: 'effectiveness'
         },
         schema,
         0
--- a/test/ranking/ranking.test.ts
+++ b/test/ranking/ranking.test.ts
@@ -1,0 +1,36 @@
+import {Channel} from 'vega-lite/src/channel';
+import {Type} from 'vega-lite/src/type';
+
+import {schema} from '../fixture';
+import {rank} from '../../src/ranking/ranking';
+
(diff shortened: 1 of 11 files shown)


kanitw:
Missing test

test/ranking/ranking.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import {rank} from '../../src/ranking/ranking';
+
+import {SpecQueryModelGroup} from '../../src/modelgroup';
+
+import {assert} from 'chai';
+
+
+describe('ranking', () => {
+  describe('rank', () => {
+    describe('group.items', () => {
```
felixcodes: @kanitw: I'm not sure if I've nested these describes properly. Do you have any comments?

kanitw: rank is the method you're testing.  

So `group.items` shouldn't be a describe instead put all of these inside `it('should return a empty group if input group is empty')`   (No need for sub-`it` below.)


(Basically group is empty is it has no items.)

PR #187, opened 2016-08-07
Add post enumeration modification to generate.ts

### LINKED ISSUE #93: Change Bandwidth if cardinality is high or if it is faceted
Scale bandWidth right now are fixed.  However, with higher cardinality field, we might want to fix them. 

Block by #2

### LINKED ISSUE #101: Change Nominal Color Scale Range Based on cardinality
Use `categorical20` for color's `scale.range` instead of default undefined (implicitly `categorical10`) if cardinality > 10.

Maybe be even interpolate hue if cardinality higher than 20.

Blocked by #2

### LINKED ISSUE #176: Add Post Enumeration Modification / Transform to `generate.ts`
Instead of returning answerSet, add a logic that can transform the `specQuery` in the `specQueryModel`  (apply `map` to `answerSet` in generate.ts if one of the modifier/stylizer is included.).

(We currently have a hack for these in [`vlplot.js`](https://github.com/vega/vega-lite-ui/blob/master/src/components/vlplot/vlplot.js#L149), but it should be migrated here.) 

A few configs to be added for the transforms

- [x] `smallBandSizeForHighCardinality: {maxCardinality: number, bandSize: number}` #93
- [x] `smallBandSizeForFacet: {maxCardinality: number, bandSize: number}` #93 
- [x] `nominalScaleForHighCardinality` #101
- [ ] `xAxisOnTopForHighYCardinalityWithoutColumn`:   {maxCardinality: number}` -- require #92

---- 
(**LATER**)

- [ ] Support taking generic stylizer as input

felixcodes:
fix #176 
- add nominalColorScaleForHighCardinality config: __fix #101__
- add smallBandSizeForFacet config: fix #93
- add smallBandSizeForHighCardinality config: fix #93 
- create Stylize.ts to handle post enumeration modification ~~fix #101~~

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -86,6 +86,12 @@ export interface QueryConfig {
   preferredNominalAxis?: Channel;
   preferredFacet?: Channel;
 
+  // Not sure where these belong..
+  smallBandSizeForHighCardinality?: {maxCardinality: number, bandSize: number};
+  smallBandSizeForFacet?: {maxCardinality: number, bandSize: number};
+  nominalScaleForHighCardinality?: {maxCardinality: number};
+
+
   // Encoding Constraints
 
   maxCardinalityForCategoricalColor?: number;
@@ -149,6 +155,11 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   preferredOrdinalAxis: Channel.Y, // ordinal on y makes it easier to read.
   preferredNominalAxis: Channel.Y, // nominal on y makes it easier to read.
   preferredFacet: Channel.ROW, // row make it easier to scroll than column
+
+  // what should their default values be?
+  smallBandSizeForHighCardinality: {maxCardinality: 10, bandSize: 12},
+  smallBandSizeForFacet: {maxCardinality: 10, bandSize: 12},
+  nominalScaleForHighCardinality: {maxCardinality: 10},
 
   // Encoding Constraints -- See description inside src/constraints/encoding.ts
   maxCardinalityForCategoricalColor: 20,
--- a/src/config.ts
+++ b/src/config.ts
@@ -94,8 +94,7 @@ export interface QueryConfig {
   typeMatchesSchemaType?: boolean;
 
   // STYLIZE
-  smallBandSizeForHighCardinality?: {maxCardinality: number, bandSize: number};
-  smallBandSizeForFacet?: {maxCardinality: number, bandSize: number};
+  smallBandSizeForHighCardinalityOrFacet?: {maxCardinality: number, bandSize: number};
   nominalScaleForHighCardinality?: {maxCardinality: number};
 
   // EFFECTIVENESS PREFERENCE
@@ -162,8 +161,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   typeMatchesSchemaType: true,
 
   // STYLIZE
-  smallBandSizeForHighCardinality: {maxCardinality: 10, bandSize: 12},
-  smallBandSizeForFacet: {maxCardinality: 10, bandSize: 12},
+  smallBandSizeForHighCardinalityOrFacet: {maxCardinality: 10, bandSize: 12},
   nominalScaleForHighCardinality: {maxCardinality: 10},
 
   // RANKING PREFERENCE
--- a/src/config.ts
+++ b/src/config.ts
@@ -162,7 +162,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   // STYLIZE
   smallBandSizeForHighCardinalityOrFacet: {maxCardinality: 10, bandSize: 12},
-  nominalColorScaleForHighCardinality: {maxCardinality: 10, palette: 'categorical20'},
+  nominalColorScaleForHighCardinality: {maxCardinality: 10, palette: 'category20'},
 
   // RANKING PREFERENCE
   maxGoodCardinalityForFacet: 5, // FIXME: revise
--- a/src/config.ts
+++ b/src/config.ts
@@ -93,15 +93,14 @@ export interface QueryConfig {
   maxCardinalityForShape?: number;
   typeMatchesSchemaType?: boolean;
 
-  // Effectiveness Preference
-  maxGoodCardinalityForColor?: number; // FIXME: revise
-  maxGoodCardinalityForFacet?: number; // FIXME: revise
-
-  // POST ENUMERATION MODIFICATION
+  // STYLIZE
   smallBandSizeForHighCardinality?: {maxCardinality: number, bandSize: number};
   smallBandSizeForFacet?: {maxCardinality: number, bandSize: number};
   nominalScaleForHighCardinality?: {maxCardinality: number};
 
+  // EFFECTIVENESS PREFERENCE
+  maxGoodCardinalityForColor?: number; // FIXME: revise
+  maxGoodCardinalityForFacet?: number; // FIXME: revise
 }
 
 export const DEFAULT_QUERY_CONFIG: QueryConfig = {
@@ -162,13 +161,12 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   maxCardinalityForShape: 6,
   typeMatchesSchemaType: true,
 
-  // Ranking Preference
-
-  maxGoodCardinalityForFacet: 5, // FIXME: revise
-  maxGoodCardinalityForColor: 7, // FIXME: revise
-
-  // POST ENUMERATION MODIFICATION
+  // STYLIZE
   smallBandSizeForHighCardinality: {maxCardinality: 10, bandSize: 12},
   smallBandSizeForFacet: {maxCardinality: 10, bandSize: 12},
   nominalScaleForHighCardinality: {maxCardinality: 10},
+
+  // RANKING PREFERENCE
+  maxGoodCardinalityForFacet: 5, // FIXME: revise
+  maxGoodCardinalityForColor: 7, // FIXME: revise
 };
--- a/src/config.ts
+++ b/src/config.ts
@@ -95,7 +95,7 @@ export interface QueryConfig {
 
   // STYLIZE
   smallBandSizeForHighCardinalityOrFacet?: {maxCardinality: number, bandSize: number};
-  nominalColorScaleForHighCardinality?: {maxCardinality: number};
+  nominalColorScaleForHighCardinality?: {maxCardinality: number, palette: string};
 
   // EFFECTIVENESS PREFERENCE
   maxGoodCardinalityForColor?: number; // FIXME: revise
@@ -162,7 +162,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   // STYLIZE
   smallBandSizeForHighCardinalityOrFacet: {maxCardinality: 10, bandSize: 12},
-  nominalColorScaleForHighCardinality: {maxCardinality: 10},
+  nominalColorScaleForHighCardinality: {maxCardinality: 10, palette: 'categorical20'},
 
   // RANKING PREFERENCE
   maxGoodCardinalityForFacet: 5, // FIXME: revise
--- a/src/config.ts
+++ b/src/config.ts
@@ -95,7 +95,7 @@ export interface QueryConfig {
 
   // STYLIZE
   smallBandSizeForHighCardinalityOrFacet?: {maxCardinality: number, bandSize: number};
-  nominalScaleForHighCardinality?: {maxCardinality: number};
+  nominalColorScaleForHighCardinality?: {maxCardinality: number};
 
   // EFFECTIVENESS PREFERENCE
   maxGoodCardinalityForColor?: number; // FIXME: revise
@@ -162,7 +162,7 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   // STYLIZE
   smallBandSizeForHighCardinalityOrFacet: {maxCardinality: 10, bandSize: 12},
-  nominalScaleForHighCardinality: {maxCardinality: 10},
+  nominalColorScaleForHighCardinality: {maxCardinality: 10},
 
   // RANKING PREFERENCE
   maxGoodCardinalityForFacet: 5, // FIXME: revise
--- a/src/config.ts
+++ b/src/config.ts
@@ -86,12 +86,6 @@ export interface QueryConfig {
   preferredNominalAxis?: Channel;
   preferredFacet?: Channel;
 
-  // Not sure where these belong..
-  smallBandSizeForHighCardinality?: {maxCardinality: number, bandSize: number};
-  smallBandSizeForFacet?: {maxCardinality: number, bandSize: number};
-  nominalScaleForHighCardinality?: {maxCardinality: number};
-
-
   // Encoding Constraints
 
   maxCardinalityForCategoricalColor?: number;
@@ -102,6 +96,12 @@ export interface QueryConfig {
   // Effectiveness Preference
   maxGoodCardinalityForColor?: number; // FIXME: revise
   maxGoodCardinalityForFacet?: number; // FIXME: revise
+
+  // POST ENUMERATION MODIFICATION
+  smallBandSizeForHighCardinality?: {maxCardinality: number, bandSize: number};
+  smallBandSizeForFacet?: {maxCardinality: number, bandSize: number};
+  nominalScaleForHighCardinality?: {maxCardinality: number};
+
 }
 
 export const DEFAULT_QUERY_CONFIG: QueryConfig = {
@@ -156,11 +156,6 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   preferredNominalAxis: Channel.Y, // nominal on y makes it easier to read.
   preferredFacet: Channel.ROW, // row make it easier to scroll than column
 
-  // what should their default values be?
-  smallBandSizeForHighCardinality: {maxCardinality: 10, bandSize: 12},
-  smallBandSizeForFacet: {maxCardinality: 10, bandSize: 12},
-  nominalScaleForHighCardinality: {maxCardinality: 10},
-
   // Encoding Constraints -- See description inside src/constraints/encoding.ts
   maxCardinalityForCategoricalColor: 20,
   maxCardinalityForFacet: 10,
@@ -171,4 +166,9 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxGoodCardinalityForFacet: 5, // FIXME: revise
   maxGoodCardinalityForColor: 7, // FIXME: revise
+
+  // POST ENUMERATION MODIFICATION
+  smallBandSizeForHighCardinality: {maxCardinality: 10, bandSize: 12},
+  smallBandSizeForFacet: {maxCardinality: 10, bandSize: 12},
+  nominalScaleForHighCardinality: {maxCardinality: 10},
 };
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -23,7 +23,7 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
     }
   });
 
-  if (opt.nominalScaleForHighCardinality || opt.smallBandSizeForHighCardinalityOrFacet) {
+  if ((opt.nominalColorScaleForHighCardinality !== null) || (opt.smallBandSizeForHighCardinalityOrFacet !== null)) {
     return stylize(answerSet, schema, opt);
   }
 
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -23,5 +23,9 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
     }
   });
 
-  return stylize(answerSet, schema, opt);
+  if (opt.nominalScaleForHighCardinality || opt.smallBandSizeForHighCardinalityOrFacet) {
+    return stylize(answerSet, schema, opt);
+  }
+
+  return answerSet;
 }
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -30,8 +30,18 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
 
   let encQIndex: Dict<EncodingQuery> = {};
 
-  if (opt.smallBandSizeForHighCardinality || opt.smallBandSizeForFacet) {
-    answerSet = answerSet.map(function(specQM) {
+  answerSet = answerSet.map(function(specQM) {
+    let encodings = specQM.specQuery.encodings;
+    let encQPositionIndex: Dict<number> = {};
+
+
+    for (let i = 0; i < encodings.length; i++) {
+      let channel = encodings[i].channel as Channel;
+      encQPositionIndex[channel] = i;
+    }
+
+
+    if (opt.smallBandSizeForHighCardinality || opt.smallBandSizeForFacet) {
 
       [Channel.ROW, Channel.Y, Channel.COLUMN, Channel.X].forEach((channel) => {
         encQIndex[channel] = specQM.getEncodingQueryByChannel(channel);
@@ -44,7 +54,7 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
 
           if (yEncQ.scale === undefined) {
             specQM.setEncodingProperty(
-              specQM.getEncodingQueryIndexByChannel(Channel.Y),
+              encQPositionIndex[Channel.Y],
               Property.SCALE,
               true,
               {name: 'scale', values: [true, false]} // not sure what this EnumSpec should be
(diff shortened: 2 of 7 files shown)


src/config.ts
```diff
@@ -150,6 +156,11 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   preferredNominalAxis: Channel.Y, // nominal on y makes it easier to read.
   preferredFacet: Channel.ROW, // row make it easier to scroll than column
 
+  // what should their default values be?
```
kanitw: Why don't you look at VLUI?

src/config.ts
```diff
@@ -86,6 +86,12 @@ export interface QueryConfig {
   preferredNominalAxis?: Channel;
   preferredFacet?: Channel;
 
+  // Not sure where these belong..
```
kanitw: They are not constraints, so definitely not here.  

They are post enumeration modification, which happens after CONSTRAINTS.  
Put it below all constraints  and add `// POST ENUMERATION MODIFICATION`

src/generate.ts
```diff
@@ -22,5 +28,101 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
     }
   });
 
+  let ENCODING_QUERY_INDEX: Dict<EncodingQuery> = {};
```
kanitw: Extract methods for all these.  

In generate, you want to keep method short, keep things in the same method have same level of details. 

The upper part is very high level code that calls other methods.  So it's bad that you add low level code just right here.   This way it's hard to to targeted unit test.

src/generate.ts
```diff
@@ -22,5 +28,101 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
...
+
+  if (opt.smallBandSizeForHighCardinality || opt.smallBandSizeForFacet) {
+    answerSet = answerSet.map(function(specQM) {
```
kanitw: we use `specM` else where in the code

src/model.ts
```diff
@@ -367,6 +367,15 @@ export class SpecQueryModel {
     return undefined;
   }
 
+  public getEncodingQueryIndexByChannel(channel: Channel) {
```
kanitw: This is O(n) and you're calling it multiple times.  So this is very inefficient.  Remove this method and just create an index by reducing `specM.spec.encodings()`.

src/generate.ts
```diff
@@ -22,5 +28,101 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
     }
   });
 
+  let ENCODING_QUERY_INDEX: Dict<EncodingQuery> = {};
+
+  if (opt.smallBandSizeForHighCardinality || opt.smallBandSizeForFacet) {
```
kanitw: why only these two?

src/generate.ts
```diff
@@ -22,5 +28,101 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
...
+          }
+        }
+      }
+
+      return specQM;
+    });
+  }
+
+  if (opt.nominalScaleForHighCardinality) {
+    answerSet = answerSet.map(function(specQM) {
```
kanitw: Write this in a way that call map only once; otherwise, it's inefficient to create new array multiple times.  Inside map you consider multiple modifiers (call multiple modifier methods).

src/generate.ts
kanitw: Because of that -- create a new file called `stylize.ts`  that contains a main `stylize` method and multiple stylizer in it that get triggers based on config (I think it's better than `modifier.ts`).

kanitw: This is not a constant, I don't know why you use UPPERCASE?

kanitw: `encQIndex`

src/generate.ts
```diff
@@ -22,5 +28,101 @@ export function generate(specQ: SpecQuery, schema: Schema, opt: QueryConfig = DE
...
+        ENCODING_QUERY_INDEX[channel] = specQM.getEncodingQueryByChannel(channel);
+      });
+
+      const yEncQ = ENCODING_QUERY_INDEX[Channel.Y];
+      if (yEncQ !== undefined) {
+        if (ENCODING_QUERY_INDEX[Channel.ROW] ||
+            schema.cardinality(yEncQ) > 10) {
+
+          if (yEncQ.scale === undefined) {
+            specQM.setEncodingProperty(
```
kanitw: You already have `encQIndex` and `yEncQ`. Just use `yEncQ`, so you don't have to deal with `setEncodingProperty` (which is more like `setEncodingPropertyEnumSpec` -- maybe we can consider renaming that later).

kanitw: Like here you can just do

yEncQ.scale = yEncQ.scale || {};  
yEncQ.scale.bandSize = ...

src/config.ts
```diff
@@ -96,6 +96,12 @@ export interface QueryConfig {
   // Effectiveness Preference
   maxGoodCardinalityForColor?: number; // FIXME: revise
   maxGoodCardinalityForFacet?: number; // FIXME: revise
+
+  // POST ENUMERATION MODIFICATION
```
kanitw: We call it stylize - this is inconsistent.

src/config.ts
```diff
@@ -160,4 +166,9 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
 
   maxGoodCardinalityForFacet: 5, // FIXME: revise
   maxGoodCardinalityForColor: 7, // FIXME: revise
+
+  // POST ENUMERATION MODIFICATION
```
kanitw: stylize

src/stylize.ts
```diff
@@ -0,0 +1,74 @@
...
+import {Dict} from './util';
+
+export function stylize(answerSet: SpecQueryModel[], schema: Schema, opt: QueryConfig): SpecQueryModel[] {
+
+  answerSet = answerSet.map(function(specM) {
+    let encQIndex: Dict<EncodingQuery> = {};
+
+    if (opt.smallBandSizeForHighCardinality || opt.smallBandSizeForFacet) {
+
+      [Channel.ROW, Channel.Y, Channel.COLUMN, Channel.X].forEach((channel) => {
```
kanitw: I already mentioned that you should break sub-logic in `stylize` into multiple methods.  

Break logic in this `if` into a method.

src/stylize.ts
```diff
@@ -0,0 +1,74 @@
...
+          if (xEncQ.scale) {
+            (xEncQ.scale as ScaleQuery).bandSize = 12;
+          }
+        }
+      }
+    }
+
+    if (opt.nominalScaleForHighCardinality) {
+
+      encQIndex[Channel.COLOR] = specM.getEncodingQueryByChannel(Channel.COLOR);
```
kanitw: Same here.

src/stylize.ts
```diff
@@ -0,0 +1,74 @@
+import {Channel} from 'vega-lite/src/channel';
+import {Type} from 'vega-lite/src/type';
+
+import {QueryConfig} from './config';
+import {SpecQueryModel} from './model';
+// import {Property} from './property';
```
kanitw: Why leaving // ?

src/config.ts
```diff
@@ -96,6 +96,12 @@ export interface QueryConfig {
   // Effectiveness Preference
   maxGoodCardinalityForColor?: number; // FIXME: revise
```
kanitw: You still have a // FIXME  -- how come this is no longer a WIP?

src/config.ts
kanitw: Move this above

```
// Effectiveness Preference
```

and change Effectiveness Preference to ALL CAPS.

src/stylize.ts
```diff
@@ -0,0 +1,74 @@
...
+        encQIndex[channel] = specM.getEncodingQueryByChannel(channel);
+      });
+
+      const yEncQ = encQIndex[Channel.Y];
+      if (yEncQ !== undefined) {
+        if (encQIndex[Channel.ROW] ||
+            schema.cardinality(yEncQ) > 10) {
+
+          if (yEncQ.scale === undefined) {
+            yEncQ.scale = yEncQ.scale || {};
```
kanitw: I don't understand why do you need to add
```
 if (yEncQ.scale === undefined) 
```

You either do 
```
if (yEncQ.scale === undefined) {
  yEncQ.scale = {};
}
```

or simply 
```
yEncQ.scale = yEncQ.scale || {};
```

which I prefer because it's shorter.   

The one you're writing doesn't make sense.

src/stylize.ts
```diff
@@ -0,0 +1,74 @@
...
+          }
+
+          if (yEncQ.scale) {
```
kanitw: Why `if (yEncQ.scale)`?    You already set it above.  So this if is always true.

src/stylize.ts
felixcodes: What if yEncQ.scale was explictly set to false? In that case we would assign {} to yEncQ.scale in addition to assigning a bandSize although the scale was set to false.

kanitw: Oh good call about that.  However, you still don't need `yEncQ.scale = yEncQ.scale || {};` when scale is already `undefined`.  

Because of that I couldn't guess that you think about this from reading (distracted).  Plus, there are no comment explaining and no test case.  

Can you add comment saying why we have to explicitly check for undefined?

src/stylize.ts
kanitw: `getEncodingQueryByChannel` takes O(n) where n = # of channels and you call this 4 times.  

Use `specM.encodings()` and reduce them to create index.

src/stylize.ts
kanitw: Ok now that you think about false/null case.  We should also consider when scale / scale.bandWidth are enumSpec / have existing values as well.

kanitw:
- [ ] combine the high cardinality and facet one

src/config.ts
```diff
@@ -93,7 +93,11 @@ export interface QueryConfig {
   maxCardinalityForShape?: number;
   typeMatchesSchemaType?: boolean;
 
-  // Effectiveness Preference
+  // STYLIZE
+  smallBandSizeForHighCardinalityOrFacet?: {maxCardinality: number, bandSize: number};
+  nominalColorScaleForHighCardinality?: {maxCardinality: number};
```
kanitw: add `palette: string`, which is default to `categorical20`.  
(But user can choose other config as well.)

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+import {isEnumSpec} from './enumspec';
+import {SpecQueryModel} from './model';
+import {EncodingQuery, ScaleQuery} from './query/encoding';
+import {Schema} from './schema';
+import {Dict} from './util';
+
+export function stylize(answerSet: SpecQueryModel[], schema: Schema, opt: QueryConfig): SpecQueryModel[] {
+
+  answerSet = answerSet.map(function(specM) {
+
```
kanitw: not sure why empty line here.

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+      // yEncQ.scale has been set to false/null.
+      // This prevents us from incorrectly overriding scale and
+      // assigning a bandSize when scale is set to false
+      if (yEncQ.scale === undefined) {
+        yEncQ.scale = {};
+      }
+
+      // We do not want to assign a bandSize if scale is set to false
+      if (yEncQ.scale && !isEnumSpec(yEncQ.scale)) {
+        if (!(yEncQ.scale as ScaleQuery).bandSize && !isEnumSpec((yEncQ.scale as ScaleQuery).bandSize)) {
```
kanitw: Isn't `!(yEncQ.scale as ScaleQuery).bandSize` or
`(yEncQ.scale as ScaleQuery).bandSize === undefined`
sufficient?  

(EnumSpec !== `undefined` anyway)

If you need this, as said before, I would assign `yEncQ.scale as ScaleQuery` to a variable so you won't have too many repetitive `as` on one line.  (You don't need it here, but next time you need it don't forget this.)

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+      // yEncQ.scale = yEncQ.scale || {} to cover the case where
+      // yEncQ.scale has been set to false/null.
+      // This prevents us from incorrectly overriding scale and
+      // assigning a bandSize when scale is set to false
+      if (yEncQ.scale === undefined) {
+        yEncQ.scale = {};
+      }
+
+      // We do not want to assign a bandSize if scale is set to false
+      if (yEncQ.scale && !isEnumSpec(yEncQ.scale)) {
```
kanitw: If scale is enumSpec, but bandSize is undefined, I think it's fine to set `bandSize`.

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+
+let encQIndex: Dict<EncodingQuery> = {};
+
+export function smallBandSizeForHighCardinalityOrFacet(specM: SpecQueryModel, schema: Schema): SpecQueryModel {
+ [Channel.ROW, Channel.Y, Channel.COLUMN, Channel.X].forEach((channel) => {
+        encQIndex[channel] = specM.getEncodingQueryByChannel(channel);
+      });
+
+  const yEncQ = encQIndex[Channel.Y];
+  if (yEncQ !== undefined) {
```
kanitw: add "and yEncQ's scaleType is ORDINAL" since bandSize is only applicable for ordinal scale.  We shouldn't set bandSize to non-ordinal scale.  

(In practice, it's ignored when rendering, but this will pollute the output VL.)

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (yEncQ.scale && !isEnumSpec(yEncQ.scale)) {
+        if (!(yEncQ.scale as ScaleQuery).bandSize && !isEnumSpec((yEncQ.scale as ScaleQuery).bandSize)) {
+          (yEncQ.scale as ScaleQuery).bandSize = 12;
+        }
+      }
+    }
+  }
+
+  const xEncQ = encQIndex[Channel.X];
+  if (xEncQ !== undefined) {
```
kanitw: Same as Y

src/stylize.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (encQIndex[Channel.COLUMN] ||
+        schema.cardinality(xEncQ) > 10) {
+
+
+      if (xEncQ.scale === undefined) {
+        xEncQ.scale = {};
+      }
+
+      if (xEncQ.scale && !isEnumSpec(xEncQ.scale)) {
```
kanitw: Same as Y

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import {assert} from 'chai';
+
+describe('stylize', () => {
+  describe('smallBandSizeForHighCardinalityOrFacet', () => {
+    it('should assign a bandSize of 12 if cardinality of Y is over 10 and bandSize is not already set', () => {
+      let specM = SpecQueryModel.build({
+          mark: Mark.BAR,
+          encodings: [
+            {channel: Channel.Y, field: 'Q', scale: {}, type: Type.QUANTITATIVE}
```
kanitw: Why setting bandSize for Q?

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+      assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, 12);
+    });
+
+    it('should assign a bandSize of 12 if cardinality of Y is over 10 and bandSize is not already set', () => {
+      let specM = SpecQueryModel.build({
+          mark: Mark.BAR,
+          encodings: [
+            {channel: Channel.Y, field: 'A', scale: {}, type: Type.QUANTITATIVE},
+            {channel: Channel.ROW, field: 'A', type: Type.QUANTITATIVE}
```
kanitw: It says cardinality, but you're actually testing ROW...

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+            encodings: [
+              {channel: Channel.Y, field: 'Q', scale: false, type: Type.QUANTITATIVE}
+            ]
+          }, schema, DEFAULT_QUERY_CONFIG);
+
+        specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+        assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, undefined);
+    });
+
+    it('should not assign a bandSize if scale is an EnumSpec', () => {
```
kanitw: It should.

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+            encodings: [
+              {channel: Channel.Y, field: 'Q', scale: {name: 'scale', values: [true, false]}, type: Type.QUANTITATIVE}
+            ]
+          }, schema, DEFAULT_QUERY_CONFIG);
+
+        specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+        assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, undefined);
+    });
+
+    it('should not assign a bandSize if bandSize is an Enum Spec', () => {
```
kanitw: (This time, it should not is good! so don't flip this one.)

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      let specM = SpecQueryModel.build({
+          mark: Mark.BAR,
+          encodings: [
+            {channel: Channel.Y, field: 'Q', scale: {bandSize: {name: 'scaleBandSize', values: [17, 21]}}, type: Type.QUANTITATIVE}
```
kanitw: bandSize of Q????

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+        assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, undefined);
+    });
+
+    it('should not assign a bandSize if scale is an EnumSpec', () => {
+        let specM = SpecQueryModel.build({
+            mark: Mark.BAR,
+            encodings: [
+              {channel: Channel.Y, field: 'Q', scale: {name: 'scale', values: [true, false]}, type: Type.QUANTITATIVE}
```
kanitw: Q bandSize?????

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+      assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, 12);
+    });
+
+    it('should not assign a bandSize if scale is false', () => {
+        let specM = SpecQueryModel.build({
+            mark: Mark.BAR,
+            encodings: [
+              {channel: Channel.Y, field: 'Q', scale: false, type: Type.QUANTITATIVE}
```
kanitw: Q BandSize....

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      specM = smallBandSizeForHighCardinalityOrFacet(specM, schema);
+      assert.equal((specM.getEncodingQueryByChannel(Channel.Y).scale as ScaleQuery).bandSize, 12);
+    });
+
+    it('should assign a bandSize of 12 if cardinality of Y is over 10 and bandSize is not already set', () => {
+      let specM = SpecQueryModel.build({
+          mark: Mark.BAR,
+          encodings: [
+            {channel: Channel.Y, field: 'A', scale: {}, type: Type.QUANTITATIVE},
```
kanitw: Q

test/generate.test.ts
```diff
@@ -592,4 +611,64 @@ describe('generate', function () {
...
+
+    it('should output bandSize = 12', () => {
+      const specQ = {
+        mark: Mark.BAR,
+        encodings: [
+          {
+            channel: Channel.Y,
+            field: 'A',
+            scale: {},
+            type: Type.QUANTITATIVE
```
kanitw: Q

test/generate.test.ts
```diff
@@ -592,4 +611,64 @@ describe('generate', function () {
...
+  describe('smallBandSizeForHighCardinalityOrFacet', () => {
+    it('should output bandSize = 12', () => {
+      const specQ = {
+        mark: Mark.BAR,
+        encodings: [
+          {
+            channel: Channel.Y,
+            field: 'Q',
+            scale: {},
+            type: Type.QUANTITATIVE
```
kanitw: Q

test/generate.test.ts
```diff
@@ -592,4 +611,64 @@ describe('generate', function () {
       });
     });
   });
+
+  describe('nominalColorScaleForHighCardinality', () => {
```
kanitw: How about wrap these two `describe` under `describe('stylizer')`.

(Keep `describe('nominalColorScaleForHighCardinality', ...)` as nested under `stylizer` though)

And move the generic case above to here.  (It `should ... when all stylizer are disabled`.)

test/generate.test.ts
```diff
@@ -19,6 +19,25 @@ const CONFIG_WITH_AUTO_ADD_COUNT = extend({}, DEFAULT_QUERY_CONFIG, {autoAddCoun
 
 describe('generate', function () {
   describe('1D', () => {
+    it('should generate answerSet with nominalColorScaleForHighCardinality and smallBandSizeForHighCardinalityOrFacet turned off', () => {
```
kanitw: This doesn't belong here.  You're trying to test stylizer, not generic 1D logic.  

Move it to below ([here](https://github.com/uwdata/compassql/pull/187/files#r73961340)).

src/stylize.ts
felixcodes: Shouldn't that be on the line with if(yEncQ.scale) since we aren't able to evaluate 
(yEncQ.scale as ScaleQuery).type in the event that yEncQ.scale is undefined? (can't get property type of undefined). I've put it on lines 49 and 67 in my most recent push.

test/stylize.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import {DEFAULT_QUERY_CONFIG} from '../src/config';
+import {SpecQueryModel} from '../src/model';
+import {ScaleQuery} from '../src/query/encoding';
+import {smallBandSizeForHighCardinalityOrFacet, nominalColorScaleForHighCardinality} from '../src/stylize';
+
+import {assert} from 'chai';
+
+describe('stylize', () => {
+  describe('smallBandSizeForHighCardinalityOrFacet', () => {
+    it('should assign a bandSize of 12 if cardinality of Y is over 10 and bandSize is not already set', () => {
```
kanitw: Don't you have to add one case when it's already set?

src/stylize.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return specM;
+}
+
+export function nominalColorScaleForHighCardinality(specM: SpecQueryModel, schema: Schema): SpecQueryModel {
+  encQIndex[Channel.COLOR] = specM.getEncodingQueryByChannel(Channel.COLOR);
+
+  const colorEncQ = encQIndex[Channel.COLOR];
+  if ((colorEncQ !== undefined) && (colorEncQ.type === Type.NOMINAL) &&
```
kanitw: why `10`?   Use value in the config.

src/stylize.ts
```diff
@@ -0,0 +1,99 @@
...
+      (schema.cardinality(colorEncQ) > 10)) {
+
+    if (colorEncQ.scale === undefined) {
+      colorEncQ.scale = {};
+    }
+
+    if (colorEncQ.scale) {
+      if (!(colorEncQ.scale as ScaleQuery).range) {
+        (colorEncQ.scale as ScaleQuery).range = 'category20';
```
kanitw: Use value in the config.

test/stylize.test.ts
```diff
@@ -0,0 +1,164 @@
...
+            encodings: [
+              {channel: Channel.COLOR, field: 'N20', scale: {name: 'scale', values: [true, false]}, type: Type.NOMINAL}
+            ]
+          }, schema, DEFAULT_QUERY_CONFIG);
+
+        specM = nominalColorScaleForHighCardinality(specM, schema);
+        assert.equal((specM.getEncodingQueryByChannel(Channel.COLOR).scale as ScaleQuery).range, 'category20');
+    });
+
+    it('should not assign a range if cardinality of color is over 10 and scale.range is an Enum Spec', () => {
```
kanitw: 20

src/stylize.ts
```diff
@@ -0,0 +1,99 @@
...
+          (yEncQ.scale as ScaleQuery).bandSize = 12;
+        }
+      }
+    }
+  }
+
+  const xEncQ = encQIndex[Channel.X];
+  if (xEncQ !== undefined) {
+    if (encQIndex[Channel.COLUMN] ||
+        schema.cardinality(xEncQ) > 10) {
```
kanitw: value in the config

test/stylize.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    it('should assign a range of category20 if cardinality of color is over 10 and range is not already set', () => {
+        let specM = SpecQueryModel.build({
+            mark: Mark.POINT,
+            encodings: [
+              {channel: Channel.COLOR, field: 'N20', scale: {}, type: Type.NOMINAL}
+            ]
+          }, schema, DEFAULT_QUERY_CONFIG);
+
+        specM = nominalColorScaleForHighCardinality(specM, schema);
+        assert.equal((specM.getEncodingQueryByChannel(Channel.COLOR).scale as ScaleQuery).range, 'category20');
```
felixcodes: @kanitw this line^

PR #197, opened 2016-08-09
add aggregateOnly and rawOnly config and constraint

### LINKED ISSUE #6: Constraint: AggregateOnly / RawOnly
- [x] AggregateOnly
- [x] RawOnly

felixcodes:
fix #6

- add aggregateOnly config
- add rawOnly config
- add aggregateOnly constraint
- add rawOnly constraint
- add tests for both constraints

### DIFF
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -102,21 +102,6 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
-    name: 'aggregateOnly',
-    description: 'Raw data should not be used if aggregateOnly config is turned on',
-    properties: [Property.AGGREGATE, Property.AUTOCOUNT],
-    allowEnumSpecForProperties: true,
-    strict: false,
-    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
-      if (opt.aggregateOnly) {
-        if (!specM.isAggregate()) {
-          return false;
-        }
-      }
-      return true;
-    }
-  },
-  {
     name: 'alwaysIncludeZeroInScaleWithBarMark',
     description: 'Do not reccommend bar mark if scale does not start at zero',
     properties: [Property.MARK, Property.SCALE, Property.SCALE_ZERO, Property.CHANNEL, Property.TYPE],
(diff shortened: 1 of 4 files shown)


felixcodes:
Travis is failing due to a network error.

src/constraint/spec.ts
```diff
@@ -102,6 +102,21 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'aggregateOnly',
+    description: 'Omit raw plots.',
+    properties: [Property.AGGREGATE, Property.AUTOCOUNT],
+    allowEnumSpecForProperties: true,
+    strict: false,
+    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      if (opt.aggregateOnly) {
```
kanitw: You don't need to check for this.  Our generation would trigger this constraints only if it is true.

src/constraint/spec.ts
```diff
@@ -429,6 +444,21 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
     }
   },
+  {
+    name: 'rawOnly',
+    description: 'Omit aggregate plots.',
+    properties: [Property.AGGREGATE, Property.AUTOCOUNT],
+    allowEnumSpecForProperties: true,
+    strict: false,
+    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      if (opt.rawOnly) {
```
kanitw: You don't need to check for this.  Our generation would trigger this constraints only if it is true.

test/constraint/spec.test.ts
```diff
@@ -33,6 +33,30 @@ describe('constraints/spec', () => {
     }
   });
 
+  describe('aggregateOnly', () => {
+    const CONFIG_WITH_AGGREGATE_ONLY = extend({}, DEFAULT_QUERY_CONFIG, {aggregateOnly: true});
+
+    it('should return false if there is raw data when aggregateOnly config is turned on', () => {
```
kanitw: cut "when", because it doesn't matter for constraint test... same for others...

kanitw:
I suggest that you open vega-editor and choose 1D and add omitRaw/Aggregate and see if the output makes sense.  If so, just copy the spec and assert the sensible result.  Should be easy to write. ;)

PR #201, opened 2016-08-10
OmitAggregatePlotWithoutDimension constraint

### LINKED ISSUE #192: Omit Aggregate Plot Without Dimension
`omitAggregatePlotWithoutDimension`

We want to be able to filter out MEAN only plots, for example. 

vega_editor

felixcodes:
Fix #192

- add OmitAggregatePlotWithoutDimension config
- add OmitAggregatePlotWithoutDimension constraint
- add Query test to verify that result excludes aggregate plot without dimension
- add unit tests for constraint


test/constraint/spec.test.ts
```diff
@@ -532,6 +532,39 @@ describe('constraints/spec', () => {
...
+        ]
+      });
+      assert.isFalse(SPEC_CONSTRAINT_INDEX['omitAggregatePlotWithoutDimension'].satisfy(specM, schema, CONFIG_WITH_OMIT_AGGREGATE_PLOT_WITHOUT_DIMENSION));
+    });
+
+    it('should return true if plot is aggregate and has dimension', () => {
+      const specM = buildSpecQueryModel({
+        mark: Mark.POINT,
+        encodings: [
+          {channel: Channel.X, field: 'N', type: Type.NOMINAL, aggregate: AggregateOp.MEAN}
```
kanitw: This is cheating.  You can't aggregate nominal in practice.  Please re-write a proper test case with N and MEAN(Q)

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
...
+
+describe('2D aggregate', () => {
+  describe('summary with mean', () => {
+    it('should not generate 2D summary plot with MEAN(Q) x MEAN(Q1)', () => {
+      const q = {
+        spec: {
+          mark: Mark.POINT,
+          encodings:[
+            {
+            channel: Channel.X,
```
kanitw: bad tabbing

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
...
+    });
+  });
+
+    it('should generate 2D summary plot with MEAN(Q) x MEAN(Q1)', () => {
+      const q = {
+        spec: {
+          mark: Mark.POINT,
+          encodings:[
+            {
+            channel: Channel.X,
```
kanitw: tab

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
     });
   });
 });
+
+describe('2D aggregate', () => {
```
kanitw: This is calling `query()` method, why don't you put this `2D` inside `query()`.  We already talked about that you should put things under namespace -> method -> cases.  

So here is a rough sketch how you can structure it: 
```
describe("query/query", ...
  describe("query", ... // (maybe rename to `query()`) 
     describe("omitAggregatePlotWithoutDimension",     // This is the config you wanna test 
        it("?(Q) x ?(Q) should not produce MEAN(Q)xMEAN(Q) if omitAggregatePlotWithoutDimension is enabled.", ...
        it("?(Q) x ?(Q) should produce MEAN(Q)xMEAN(Q) if omitAggregatePlotWithoutDimension is disabled.", ... 
```

With this good hierarchy, you don't even need to add comment at the end.

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
...
+  describe('summary with mean', () => {
+    it('should not generate 2D summary plot with MEAN(Q) x MEAN(Q1)', () => {
```
kanitw: Your test case description is still bad.  This doesn't explain why this one should not generate and why next one should.

src/constraint/spec.ts
```diff
@@ -264,6 +265,24 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'omitAggregatePlotWithoutDimension',
+    description: 'Aggregate plots without dimension should be omitted',
+    properties: [Property.AGGREGATE, Property.AUTOCOUNT, Property.TYPE],
+    allowEnumSpecForProperties: false,
```
kanitw: Why should this one be `false`?

kanitw: I'm not 100% sure, but wanna hear your explanation.

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
...
+        },
+        nest: [{
+          groupBy: [Property.FIELD, Property.AGGREGATE, Property.BIN, Property.TIMEUNIT]
+        }],
+        config: {
+          autoAddCount: true,
+          omitAggregatePlotWithoutDimension: false
+        }
+      };
+      const CONFIG_WITH_OMIT_AGGREGATE_PLOT_WITHOUT_DIMENSION = extend({}, DEFAULT_QUERY_CONFIG, {omitAggregatePlotWithoutDimension: false});
```
kanitw: I don't understand why you need to add this, which is not only unnecessary but also increase the chance that you'll forget things.

test/query/query.test.ts
```diff
@@ -97,3 +99,77 @@ describe('query/query', () => {
...
+        },
+        nest: [{
+          groupBy: [Property.FIELD, Property.AGGREGATE, Property.BIN, Property.TIMEUNIT]
+        }],
+        config: {
+          autoAddCount: true,
+          omitAggregatePlotWithoutDimension: true
+        }
+      };
+      const CONFIG_WITH_OMIT_AGGREGATE_PLOT_WITHOUT_DIMENSION = extend({}, DEFAULT_QUERY_CONFIG, {omitAggregatePlotWithoutDimension: true});
```
kanitw: I don't understand why you need to add this config, which is not only unnecessary but also increase the chance that you'll forget things.

test/constraint/spec.test.ts
```diff
@@ -532,6 +532,40 @@ describe('constraints/spec', () => {
     });
   });
 
+  describe('omitAggregatePlotWithoutDimension', () => {
+    const CONFIG_WITH_OMIT_AGGREGATE_PLOT_WITHOUT_DIMENSION = extend({}, DEFAULT_QUERY_CONFIG, {omitAggregatePlotWithoutDimension: true});
```
kanitw: Why do you need this?

kanitw: Please also remove similar things from your old `omitRaw` and `omitAggregate` tests.

src/constraint/spec.ts
```diff
@@ -264,6 +265,24 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'omitAggregatePlotWithoutDimension',
+    description: 'Aggregate plots without dimension should be omitted',
+    properties: [Property.AGGREGATE, Property.AUTOCOUNT, Property.TYPE],
```
kanitw: Are these sufficient for checking `isDimension`?   I bet you haven't checked that.

kanitw: So everytime you're not sure about something you should investigate, not just guess.  
I almost didn't catch this.   Don't expect that I'll catch every single line.

PR #207, opened 2016-08-10
[Follow-up]  Test for #199 and revert `omitAggregate` to allow enum spec

### LINKED ISSUE #199: This Spec should also returns histogram
```
{
  "spec": {
    "mark": "?",
    "encodings": [
      {
        "channel": "?",
        "bin": "?",
        "aggregate": "?",
        "field": "Acceleration",
        "type": "?"
      }
    ]
  },
  "groupBy": ["field", "aggregate", "bin", "timeUnit"],
  "chooseBy": "effectiveness",
  "config": {
          "autoAddCount": true,
          "omitRaw": true
        }
}
```

This only returns a mean plot.

vega_editor

felixcodes:
fix #199 
- Add test: do not exclude autocount if it is still enumspec for omitRaw constraint
- Add test: do not exclude aggregate if it is still enumspec for omitRaw constraint
- set allowEnumSpecForProperties: true for omitAggregate constraint

### DIFF
--- a/test/constraint/spec.test.ts
+++ b/test/constraint/spec.test.ts
@@ -858,7 +858,7 @@ describe('constraints/spec', () => {
       assert.isFalse(SPEC_CONSTRAINT_INDEX['omitRaw'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
     });
 
-    it('should not exclude aggregate if aggregate if it is still an enum spec', () => {
+    it('should not exclude aggregate if it is still an enum spec', () => {
       const specM = buildSpecQueryModel({
         mark: Mark.POINT,
         encodings: [
--- a/test/constraint/spec.test.ts
+++ b/test/constraint/spec.test.ts
@@ -858,7 +858,7 @@ describe('constraints/spec', () => {
       assert.isFalse(SPEC_CONSTRAINT_INDEX['omitRaw'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
     });
 
-    it('should not exclude aggregate if it is still an enum spec', () => {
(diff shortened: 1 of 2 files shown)


test/constraint/spec.test.ts
```diff
@@ -858,6 +858,26 @@ describe('constraints/spec', () => {
       assert.isFalse(SPEC_CONSTRAINT_INDEX['omitRaw'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
     });
 
+    it('should not exclude aggregate if it is still an enum spec', () => {
```
kanitw: What does "exclude aggregate" mean?  It reads very confusing.  

The method returns true or false.  Why don't you use consistent message as others?
e.g., `'It should return true if there is an enum spec aggregate'`.
 
(As mentioned this morning, you definitely need to spend more time with test description as I have been correcting them for you for almost every PRs.)

test/constraint/spec.test.ts
```diff
@@ -858,6 +858,26 @@ describe('constraints/spec', () => {
...
+      const specM = buildSpecQueryModel({
+        mark: Mark.POINT,
+        encodings: [
+          {aggregate: SHORT_ENUM_SPEC, channel: Channel.X, field: 'A', type: Type.QUANTITATIVE}
+        ]
+      });
+      assert.isTrue(SPEC_CONSTRAINT_INDEX['omitRaw'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
+    });
+
+    it('should not exclude autocount if it is still an enum spec', () => {
```
kanitw: Same here.

PR #211, opened 2016-08-10
Convert SpecQueryModelGroup to Class

### LINKED ISSUE #181: Convert SpecQueryModelGroup to Class
So that people won't modify it's inner context randomly. 
Plus, `getTopItem` can be a class method.

felixcodes:
fix #181 
- add SpecQueryModelGroup class in model.ts
- Remove `cql.modelGroup.isSpecQueryModelGroup` method (we can reliably use instanceof)
- `cql.modelGroup.getTopItem` to `SpecQueryModelGroup.getTopSpecQueryModel`


src/model.ts
```diff
@@ -462,3 +463,61 @@ export class SpecQueryModel {
...
+
+  public set items(items: (SpecQueryModel | SpecQueryModelGroup)[]) {
+    this._items = items;
+  }
+
+  public set groupBy(groupBy: GroupBy) {
+    this._groupBy = groupBy;
+  }
+
+  // Currently, we do not need to get orderGroupBy for anything in CQL.
```
felixcodes: @kanitw

src/model.ts
```diff
@@ -462,3 +463,61 @@ export class SpecQueryModel {
...
+  constructor(name: string = '', path: string = '', items: (SpecQueryModel | SpecQueryModelGroup)[] = [],
+              groupBy: GroupBy = undefined, orderGroupBy: string = undefined) {
+    this._name = name;
+    this._path = path;
+    this._items = items;
+    this.groupBy = groupBy;
+    this._orderGroupBy = orderGroupBy;
+  }
+
+  public getTopSpecModel(g: SpecQueryModelGroup): SpecQueryModel {
```
kanitw: getTopSpecQueryModel

src/model.ts
kanitw: Have you check if we use them in editor / vlui? 

It's kinda weird if you can set something but can't get them.  Usually it's the other way around.  (Can `get` but not `set`)

felixcodes: still in progress of looking at other files

src/model.ts
```diff
@@ -462,3 +463,61 @@ export class SpecQueryModel {
...
+    } else {
+      return topItem;
+    }
+  }
+
+  public get name() {
+    return this._name;
+  }
+
+  public set name(name: string) {
```
kanitw: Do we need all of these setters?  Is there way to refactor code such that a lot of these things get defined when the constructor is called? 

Generally, I feel like if we have both getter and setters for everything, why don't we make all of them just a public members?

src/model.ts
```diff
@@ -462,3 +463,53 @@ export class SpecQueryModel {
...
+  constructor(name: string = '', path: string = '', items: (SpecQueryModel | SpecQueryModelGroup)[] = [],
+              groupBy: GroupBy = undefined, orderGroupBy: string = undefined) {
+    this._name = name;
+    this._path = path;
+    this._items = items;
+    this._groupBy = groupBy;
+    this._orderGroupBy = orderGroupBy;
+  }
+
+  public getTopSpecQueryModel(g: SpecQueryModelGroup): SpecQueryModel {
```
kanitw: Why should this accept g?   

It's now a class method.  Just use `this` instead of `g`.

PR #213, opened 2016-08-11
[WIP] Add shorthand support for transform

### LINKED ISSUE #171: Add shorthand support for transform
filter: 
string - keep as is
filter object - there's a method to call method to translate filter into expression. use it, turn it into expression

felixcodes:
fix #171 
- Add shorthand support for transform
- Add tests for transform shorthand
- rewrite stack to not use template strings

### DIFF
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -34,7 +34,7 @@ export function value(v: any, replace: Replacer): any {
 
 export const INCLUDE_ALL: Dict<boolean> =
   // TODO: remove manual stack concat once we really support enumerating it.
-  DEFAULT_PROPERTY_PRECEDENCE.concat([Property.STACK])
+  DEFAULT_PROPERTY_PRECEDENCE.concat([Property.STACK, Property.FILTER, Property.CALCULATE])
     .reduce((m, prop) => {
       m[prop] = true;
       return m;
@@ -57,12 +57,16 @@ export function spec(specQ: SpecQuery,
   }
 
   if (specQ.transform) {
-    if (specQ.transform.calculate !== undefined) {
-      parts.push(`calculate:${calculate(specQ.transform.calculate)}`);
+    if (include[Property.CALCULATE]) {
+      if (specQ.transform.calculate !== undefined) {
+        parts.push('calculate:' + calculate(specQ.transform.calculate));
+      }
     }
 
-    if (specQ.transform.filter !== undefined) {
-      parts.push(`filter:${filter(specQ.transform.filter)}`);
+    if (include[Property.FILTER]) {
+      if (specQ.transform.filter !== undefined) {
+        parts.push('filter:' + filter(specQ.transform.filter));
+      }
     }
   }
 
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -1,3 +1,6 @@
+import {expression} from 'vega-lite/src/filter';
+import {Filter} from 'vega-lite/src/filter';
+import {Formula} from 'vega-lite/src/transform';
 import {Type} from 'vega-lite/src/type';
 import {isString} from 'datalib/src/util';
 
@@ -53,7 +56,15 @@ export function spec(specQ: SpecQuery,
     parts.push(value(specQ.mark, replace[Property.MARK]));
   }
 
-  // TODO: transform
+  if (specQ.transform) {
+    if (specQ.transform.calculate !== undefined) {
+      parts.push(`calculate:${calculate(specQ.transform.calculate)}`);
+    }
+
+    if (specQ.transform.filter !== undefined) {
+      parts.push(`filter:${filter(specQ.transform.filter)}`);
+    }
+  }
 
   // TODO: extract this to its own stack method
   if (include[Property.STACK]) {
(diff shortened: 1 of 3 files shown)


src/query/shorthand.ts
```diff
@@ -53,7 +56,13 @@ export function spec(specQ: SpecQuery,
     parts.push(value(specQ.mark, replace[Property.MARK]));
   }
 
-  // TODO: transform
+  if (specQ.transform) {
+    const calc = stringifyCalculate(specQ.transform.calculate);
+    const filter = stringifyFilter(specQ.transform.filter);
+
+    parts.push((calc ? `calculate:${calc}` : '') +
```
felixcodes: @kanitw I'm not sure what the best way to divide calculate and and filter would be, especially since there are occasions where one or the other may not appear. Should we leave it as is, or do you have a specific delimiter in mind?

src/query/shorthand.ts
```diff
@@ -91,6 +100,54 @@ export function spec(specQ: SpecQuery,
   return parts.join('|');
 }
 
+let getFilterExpression = filter.getFilterExpression;
+
+export function stringifyFilter(filter: string | Filter | (string | Filter)[]): string {
```
felixcodes: should this method simply be named `filter` instead?

src/query/shorthand.ts
```diff
@@ -91,6 +100,54 @@ export function spec(specQ: SpecQuery,
...
+  } else if (typeof filter === 'string') {
+    filterExpression += filter;
+  } else { // FilterObj
+    filterExpression += getFilterExpression(filter);
+  }
+
+  return filterExpression;
+}
+
+export function stringifyCalculate(formulaArr: Formula[]): string {
```
felixcodes: @kanitw  should this method be named calculate instead?

src/query/shorthand.ts
```diff
@@ -91,6 +100,54 @@ export function spec(specQ: SpecQuery,
   return parts.join('|');
 }
 
+let getFilterExpression = filter.getFilterExpression;
```
felixcodes: I'm considering changing `filter` to filterNameSpace so I can change the name of the method below.

felixcodes:
The case that is failing in Travis passes on my machine. Will look into it tomorrow morning.

src/query/shorthand.ts
```diff
@@ -1,4 +1,7 @@
 import {Type} from 'vega-lite/src/type';
+import {filter} from 'vega-lite/src/compile/data/filter';
```
kanitw: You supposed to move it to `src/filter`  https://github.com/vega/vega-lite/issues/1491

src/query/shorthand.ts
kanitw: As we promote them to top-level n shorthand.  Why don't you consider these as different parts that get pushed separately.  (Thus get  top-level `|` delimiter automatically.)

src/query/shorthand.ts
kanitw: If you have done this https://github.com/vega/vega-lite/issues/1491, you won't have this problem.

src/query/shorthand.ts
kanitw: yes

src/query/shorthand.ts
kanitw: Plus, don't push empty string if a particular part is empty.  Otherwise you get `||`

src/query/shorthand.ts
```diff
@@ -91,6 +102,45 @@ export function spec(specQ: SpecQuery,
   return parts.join('|');
 }
 
+export function filter(_filter: string | Filter | (string | Filter)[]): string {
+  let filterExpression = '';
+
+  if (_filter instanceof Array) {
+    if (_filter.length > 0) {
+      if (typeof _filter[0] === 'string') {
```
kanitw: use `isString` in util

kanitw: Your logic here doesn't allow mixing up `string`  and object. 
Why don't you do just a simple map and check type inside the map and just join at the end. 

Plus, I think `) && (` is a better delimiter with `(` and `)` at the beginning and end.

src/query/shorthand.ts
```diff
@@ -91,6 +102,45 @@ export function spec(specQ: SpecQuery,
   return parts.join('|');
 }
 
+export function filter(_filter: string | Filter | (string | Filter)[]): string {
+  let filterExpression = '';
+
+  if (_filter instanceof Array) {
+    if (_filter.length > 0) {
```
kanitw: If you use map, no need to check length.

src/query/shorthand.ts
```diff
@@ -91,6 +102,45 @@ export function spec(specQ: SpecQuery,
...
+    filterExpression += _filter;
+  } else { // FilterObj
+    filterExpression += expression(_filter);
+  }
+
+  return filterExpression;
+}
+
+export function calculate(formulaArr: Formula[]): string {
+  if (formulaArr.length > 0) {
```
kanitw: If you use map, no need to check length.

src/query/shorthand.ts
```diff
@@ -53,7 +56,15 @@ export function spec(specQ: SpecQuery,
     parts.push(value(specQ.mark, replace[Property.MARK]));
   }
 
-  // TODO: transform
+  if (specQ.transform) {
```
kanitw: you haven't take `include` into account at all.

PR #224, opened 2016-08-16
Cardinality exclude invalid flag

RileyChang:
1) Add a third argument, `excludeInvalid`, to schema's `cardinality` method that causes invalid values to be ignored when calculating the cardinality. Ignored values are: `null`, `NaN`, and any invalid date in the case of temporal data.

2) Updates the `timeUnitShouldShowVariation` constraint to be more accurate by using the change in (1)

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -94,7 +94,7 @@ export class Schema {
     for (let fieldSchema of fieldSchemas) {
       if (fieldSchema.type === Type.QUANTITATIVE) {
         for (let maxbins of opt.maxBinsList) {
-          fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats, false);
+          fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
         }
       } else if (fieldSchema.type === Type.TEMPORAL) {
         for (let unit of opt.timeUnits) {
@@ -155,8 +155,9 @@ export class Schema {
       const maxbins: any = bin.maxbins;
       if (!fieldSchema.binStats[maxbins]) {
         // need to calculate
-        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats, excludeInvalid);
+        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
       }
+      // don't need to worry about excludeInvalid here because invalid values don't affect linearly binned field's cardinality
       return fieldSchema.binStats[maxbins].distinct;
     } else if (encQ.timeUnit) {
       if (augmentTimeUnitDomain) {
@@ -225,7 +226,7 @@ export class Schema {
 /**
  * @return a summary of the binning scheme determined from the given max number of bins
  */
-function binSummary(maxbins: number, summary: Summary, excludeInvalid: boolean): Summary {
+function binSummary(maxbins: number, summary: Summary): Summary {
   const bin = dlBin({
     min: summary.min,
     max: summary.max,
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -99,7 +99,7 @@ export class Schema {
       } else if (fieldSchema.type === Type.TEMPORAL) {
         for (let unit of opt.timeUnits) {
           if (unit !== undefined) {
-            fieldSchema.timeStats[unit] = timeSummary(unit, fieldSchema.stats, false);
+            fieldSchema.timeStats[unit] = timeSummary(unit, fieldSchema.stats);
           }
         }
       }
@@ -176,12 +176,14 @@ export class Schema {
       var timeStats = fieldSchema.timeStats;
       // if the cardinality for the timeUnit is not cached, calculate it
       if (!timeStats[unit]) {
-        timeStats[unit] = timeSummary(encQ.timeUnit as TimeUnit, fieldSchema.stats, excludeInvalid);
-      } else if (excludeInvalid) {
-        // we have it cached, but we have to make sure it excludes invalid values
-        return timeStats[unit].distinct - invalidCount(timeStats[unit].unique, ['Invalid Date']);
+        timeStats[unit] = timeSummary(encQ.timeUnit as TimeUnit, fieldSchema.stats);
       }
-      return timeStats[unit].distinct;
+
+      if (excludeInvalid) {
+        return timeStats[unit].distinct - invalidCount(timeStats[unit].unique, ['Invalid Date', null]);
+      } else {
+        return timeStats[unit].distinct;
+      }
     } else {
       if (fieldSchema) {
         if (excludeInvalid) {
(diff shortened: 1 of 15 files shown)


src/schema.ts
```diff
@@ -155,7 +155,7 @@ export class Schema {
       const maxbins: any = bin.maxbins;
       if (!fieldSchema.binStats[maxbins]) {
         // need to calculate
-        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
+        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats, false);
```
kanitw: Looks like all occurrences of `binSummary` call this with `excludeInvalid=false` ignoring the input `excludeInvalid`.    I don't understand why it has to always be false.  Plus, if you don't wanna use it, why adding this parameter to `binSummary` method at all?

src/schema.ts
```diff
@@ -228,14 +231,17 @@ function binSummary(maxbins: number, summary: Summary): Summary {
...
-function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
+function timeSummary(timeunit: TimeUnit, summary: Summary, excludeInvalid: boolean): Summary {
   const result = extend({}, summary);
 
   var unique: {[value: string]: number} = {};
   keys(summary.unique).forEach(function(dateString) {
-    let date: Date = new Date(dateString);
-    let key: string = ((timeunit === TimeUnit.DAY) ? date.getDay() : convert(timeunit, new Date(dateString))).toString();
-    unique[key] = unique[key] ? unique[key] + summary.unique[dateString] : summary.unique[dateString];
+    // only add key to unique if excludeInvalid is not set or if it is NOT undefined, null, or NaN
```
kanitw: > only add key to unique if excludeInvalid is not set or if it is NOT undefined, null, or NaN

This will make the cache data incorrect if the first time you ask for `excludeInvalid = true` and the next time you ask for false.  

We discussed earlier today that we should always add these invalid values to unique dictionary but use `unique` as a part of cardinality calculation.  I'm not sure why you change it like this and it seems incorrect.

src/schema.ts
```diff
@@ -247,19 +253,33 @@ function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
...
+      } else {
+        newUnique[bucket] += oldUnique[value];
+      }
     }
   }
   return newUnique;
 }
 
+// returns the number of distinct values not including values in list
+function distinctExcluding(stats: Summary, list: string[]): number {
```
kanitw: This is expensive.  Instead of running through all key in unique.  You should just have an `invalidCount` method that takes `unique` dictionary (instead of the whole summary, which is more than you need) and `list` of invalid values.  Then return numbers of values in the list and you can use this number to be subtracted from the full distinct (possibly cached).   Note that you always cache the full distinct and you can easily call this `invalidCount` to subtract every time.

src/schema.ts
```diff
@@ -174,11 +174,14 @@ export class Schema {
       }
       // if the cardinality for the timeUnit is not cached, calculate it
       if (!fieldSchema.timeStats[encQ.timeUnit as string]) {
-        fieldSchema.timeStats[encQ.timeUnit as string] = timeSummary(encQ.timeUnit as TimeUnit, fieldSchema.stats);
+        fieldSchema.timeStats[encQ.timeUnit as string] = timeSummary(encQ.timeUnit as TimeUnit, fieldSchema.stats, excludeInvalid);
```
kanitw: As said [here](https://github.com/uwdata/compassql/pull/224/files#r74876060), you should always cache the full distinct and subtract `invalidCount()` if the flag is enabled.

src/schema.ts
```diff
@@ -174,11 +174,14 @@ export class Schema {
...
+      } else if (excludeInvalid) {
+        // we have it cached, but we have to make sure it excludes invalid values
+        return distinctExcluding(fieldSchema.timeStats[encQ.timeUnit as string], ['Invalid Date']);
```
kanitw: Why doesn't this include `null`?  Didn't we talk that `null` shouldn't even be converted in the date conversion method?

src/schema.ts
```diff
@@ -228,14 +231,17 @@ function binSummary(maxbins: number, summary: Summary): Summary {
...
   var unique: {[value: string]: number} = {};
   keys(summary.unique).forEach(function(dateString) {
-    let date: Date = new Date(dateString);
-    let key: string = ((timeunit === TimeUnit.DAY) ? date.getDay() : convert(timeunit, new Date(dateString))).toString();
-    unique[key] = unique[key] ? unique[key] + summary.unique[dateString] : summary.unique[dateString];
+    // only add key to unique if excludeInvalid is not set or if it is NOT undefined, null, or NaN
+    if (!excludeInvalid || (dateString && dateString !== 'null' && dateString !== 'NaN')) {
+      let date: Date = new Date(dateString);
+      let key: string = ((timeunit === TimeUnit.DAY) ? date.getDay() : convert(timeunit, date)).toString();
+      unique[key] = unique[key] ? unique[key] + summary.unique[dateString] : summary.unique[dateString];
```
kanitw: ` unique[key] = (unique[key] || 0) + summary.unique[dateString]; `

is quite shorter.

src/schema.ts
```diff
@@ -228,14 +231,17 @@ function binSummary(maxbins: number, summary: Summary): Summary {
...
+function timeSummary(timeunit: TimeUnit, summary: Summary, excludeInvalid: boolean): Summary {
   const result = extend({}, summary);
 
   var unique: {[value: string]: number} = {};
   keys(summary.unique).forEach(function(dateString) {
-    let date: Date = new Date(dateString);
-    let key: string = ((timeunit === TimeUnit.DAY) ? date.getDay() : convert(timeunit, new Date(dateString))).toString();
-    unique[key] = unique[key] ? unique[key] + summary.unique[dateString] : summary.unique[dateString];
+    // only add key to unique if excludeInvalid is not set or if it is NOT undefined, null, or NaN
+    if (!excludeInvalid || (dateString && dateString !== 'null' && dateString !== 'NaN')) {
```
kanitw: 1) Is dateString ever `'NaN'`?  

2) You never add `null` to the `unique` dict.

src/schema.ts
```diff
@@ -172,13 +172,28 @@ export class Schema {
...
+        return timeStats[unit].distinct - invalidCount(timeStats[unit].unique, ['Invalid Date', null]);
+      } else {
+        return timeStats[unit].distinct;
       }
-      return fieldSchema.timeStats[encQ.timeUnit as string].distinct;
     } else {
-      return fieldSchema ? fieldSchema.stats.distinct : null;
+      if (fieldSchema) {
+        if (excludeInvalid) {
+          return fieldSchema.stats.distinct - invalidCount(fieldSchema.stats.unique, ['null', 'NaN']);
```
kanitw: Why is this one `'null'` and the above is `null`.  I guess it doesn't matter but they should be consistent.

src/schema.ts
```diff
@@ -247,19 +273,32 @@ function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
 /**
  * @return a new unique object based off of the old unique count and a binning scheme
  */
-function binUnique(bin, oldUnique) {
+function binUnique(bin, oldUnique, excludeInvalid: boolean) {
```
kanitw: I think you're running into the same problem with bin here (that your cached value is inconsistent?)

Instead, why don't you always add `NaN` and `null` as separate categories in the `newUnique`

src/schema.ts
```diff
@@ -247,19 +273,32 @@ function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
...
   const newUnique = {};
   for (var value in oldUnique) {
     let bucket: number = bin.value(Number(value)) as number;
-    if (!newUnique[bucket]) {
-      newUnique[bucket] = oldUnique[value];
-    } else {
-      newUnique[bucket] += oldUnique[value];
+    if (!excludeInvalid || (bucket !== null && bucket === bucket)) {
```
kanitw: I believe we have `isNaN` method in util, which is easier to read.   
(Check NaN with `x===x`  is a javascript quirk, not every one knows it ... at least you should comment.)

src/schema.ts
```diff
@@ -247,19 +273,32 @@ function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
...
+      if (!newUnique[bucket]) {
+        newUnique[bucket] = oldUnique[value];
```
kanitw: `newUnique[bucket] = (newUnique[bucket] || 0) + oldUnique[value];`  is more concise.  :)

src/schema.ts
```diff
@@ -247,19 +273,32 @@ function timeSummary(timeunit: TimeUnit, summary: Summary): Summary {
...
+      }
     }
   }
   return newUnique;
 }
 
+/** @return the number of items in list that occur as keys of unique */
+function invalidCount(unique: {}, list: string[]) {
+  var count = 0;
+  list.forEach(function(item) {
```
kanitw: (very minor) reduce would be shorter and more concise.  :)

src/schema.ts
```diff
@@ -210,7 +225,7 @@ export class Schema {
 /**
  * @return a summary of the binning scheme determined from the given max number of bins
  */
-function binSummary(maxbins: number, summary: Summary): Summary {
+function binSummary(maxbins: number, summary: Summary, excludeInvalid: boolean): Summary {
```
kanitw: Don't add `excludeInvalid` if you're not using the parameter.

src/schema.ts
```diff
@@ -155,7 +155,7 @@ export class Schema {
       const maxbins: any = bin.maxbins;
       if (!fieldSchema.binStats[maxbins]) {
         // need to calculate
-        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats);
+        fieldSchema.binStats[maxbins] = binSummary(maxbins, fieldSchema.stats, excludeInvalid);
```
kanitw: Why bin doesn't have to subtract NaN and null like in other cases?

PR #232, opened 2016-08-16
Prevent outputting inapplicable `scale`, `sort`, `axis`, `legend` from `toSpec()` and Shorthand

### LINKED ISSUE #226: Write `toSpec()` and `toShorthand()` in a way that prevents outputting inapplicable `scale`, `sort`, `axis`/`legend`.
For example,

x/y can't have legend
detail can't have any of these

(None of the channels support all 4 -- take a look at Vega-Lite interfaces!)

felixcodes:
Fix #226
In this PR:
- Create dictionaries of channels that support `scale`, `sort`,`axis`, `legend` properties
- Prevent toSpec and Shorthand from outputting properties that are used with inapplicable channel.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -14,7 +14,7 @@ import {EnumSpecIndex} from './enumspecindex';
 import {SpecQuery, isAggregate, stack} from './query/spec';
 import {isDimension, isMeasure, EncodingQuery} from './query/encoding';
 import {GroupBy} from './query/groupby';
-import {spec as specShorthand} from './query/shorthand';
+import {spec as specShorthand, CHANNEL_SUPPORTS_AXIS, CHANNEL_SUPPORTS_LEGEND, CHANNEL_SUPPORTS_SCALE, CHANNEL_SUPPORTS_SORT} from './query/shorthand';
 import {RankingScore} from './ranking/ranking';
 import {Schema} from './schema';
 import {Dict, duplicate, extend} from './util';
@@ -694,7 +694,25 @@ export class SpecQueryModel {
 
         // otherwise, assign the proper to the field def
         if (encQ[prop] !== undefined) {
-          fieldDef[prop] = encQ[prop];
+          if (prop === Property.AXIS && !isEnumSpec(encQ.channel)) {
+            if (CHANNEL_SUPPORTS_AXIS[encQ.channel as Channel]) {
+              fieldDef[Property.AXIS] = encQ[Property.AXIS];
+            }
+          } else if (prop === Property.LEGEND && !isEnumSpec(encQ.channel)) {
+            if (CHANNEL_SUPPORTS_LEGEND[encQ.channel as Channel]) {
+              fieldDef[Property.LEGEND] = encQ[Property.LEGEND];
+            }
+          } else if (prop === Property.SCALE && !isEnumSpec(encQ.channel)) {
+            if (CHANNEL_SUPPORTS_SCALE[encQ.channel as Channel]) {
+              fieldDef[Property.SCALE] = encQ[Property.SCALE];
+            }
+          } else if (prop === Property.SORT && !isEnumSpec(encQ.channel)) {
+            if (CHANNEL_SUPPORTS_SORT[encQ.channel as Channel]) {
+              fieldDef[Property.SORT] = encQ[Property.SORT];
+            }
+          } else {
+            fieldDef[prop] = encQ[prop];
+          }
         }
       }
 
--- a/src/model.ts
+++ b/src/model.ts
@@ -695,9 +695,10 @@ export class SpecQueryModel {
         // otherwise, assign the proper to the field def
         if (encQ[prop] !== undefined) {
 
-          // We only output axis, legend, scale, or sort that have correct supporting channel
+          // axis, legend, scale, and sort only support certain channels
           let propNeedsToCheckChannel: boolean = contains([Property.AXIS, Property.LEGEND, Property.SCALE, Property.SORT], prop);
 
+          // We only output axis, legend, scale, or sort that have correct supporting channel
           if ((propNeedsToCheckChannel && !isEnumSpec(encQ.channel) && PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel]) ||
               !propNeedsToCheckChannel) {
             fieldDef[prop] = encQ[prop];
--- a/src/model.ts
+++ b/src/model.ts
@@ -17,7 +17,7 @@ import {GroupBy} from './query/groupby';
 import {spec as specShorthand, PROPERTY_SUPPORTED_CHANNELS} from './query/shorthand';
 import {RankingScore} from './ranking/ranking';
 import {Schema} from './schema';
-import {contains, Dict, duplicate, extend} from './util';
+import {Dict, duplicate, extend} from './util';
 
 export function getDefaultName(prop: Property) {
   switch (prop) {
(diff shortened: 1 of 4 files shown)


src/model.ts
```diff
@@ -694,7 +694,25 @@ export class SpecQueryModel {
 
         // otherwise, assign the proper to the field def
         if (encQ[prop] !== undefined) {
-          fieldDef[prop] = encQ[prop];
+          if (prop === Property.AXIS) {
```
kanitw: Use `switch` would be more effective here.

src/model.ts
```diff
@@ -694,7 +694,25 @@ export class SpecQueryModel {
...
+            if (CHANNEL_SUPPORTS_AXIS[encQ.channel as Channel]) {
```
kanitw: You haven't addressed my comment about enum spec?

src/query/shorthand.ts
```diff
@@ -49,6 +50,34 @@ export function vlSpec(vlspec: ExtendedUnitSpec,
   return spec(specQ);
 }
 
+export const CHANNEL_SUPPORTS_AXIS: Dict<boolean> =
+  [Channel.X, Channel.Y]
```
kanitw: ROW/COLUMN has axis too

src/query/shorthand.ts
```diff
@@ -49,6 +50,34 @@ export function vlSpec(vlspec: ExtendedUnitSpec,
...
+      m[channel] = true;
+      return m;
+    }, {} as Dict<boolean>);
+
+export const CHANNEL_SUPPORTS_SCALE: Dict<boolean> =
+  [Channel.X, Channel.Y, Channel.COLOR, Channel.OPACITY, Channel.OPACITY, Channel.SIZE, Channel.SHAPE]
+    .reduce((m, channel) => {
+      m[channel] = true;
+      return m;
+    }, {} as Dict<boolean>);
```
kanitw: ROW/COLUMN has axis too

src/query/shorthand.ts
```diff
@@ -203,10 +232,20 @@ export function fieldDef(encQ: EncodingQuery,
     }
   }
 
-  // Scale
-  // TODO(#226):
-  // write toSpec() and toShorthand() in a way that prevents outputting inapplicable scale, sort, axis / legend
   for (const nestedPropParent of [Property.SCALE, Property.SORT, Property.AXIS, Property.LEGEND]) {
+    if ((nestedPropParent === Property.AXIS) && !isEnumSpec(encQ.channel) && !CHANNEL_SUPPORTS_AXIS[encQ.channel as Channel]) {
```
kanitw: These 4 conditions seems unnecessary redundant.  Can't you combine them to one line?

src/model.ts
felixcodes: https://github.com/uwdata/compassql/blob/fo/226-tospec-shorthand-output/src/model.ts#L683

src/model.ts
kanitw: change 

CHANNEL_SUPPORTS_LEGEND[encQ.channel as Channel]

to 

CHANNEL_SUPPORTS_PROPERTY[channel]['legend' / 'scale'/ ...]

and your code will be way shorter

src/query/shorthand.ts
kanitw: change 
```
CHANNEL_SUPPORTS_LEGEND[encQ.channel as Channel]
```
to 
```
CHANNEL_SUPPORTS_PROPERTY[channel]['legend' / 'scale'/ ...]
```
and your code will be way shorter

felixcodes: Dict<Dict<boolean>>

src/model.ts
```diff
@@ -694,7 +694,14 @@ export class SpecQueryModel {
 
         // otherwise, assign the proper to the field def
         if (encQ[prop] !== undefined) {
-          fieldDef[prop] = encQ[prop];
+
+          if (prop === Property.AXIS || prop === Property.LEGEND || prop === Property.SCALE || prop === Property.SORT) {
```
kanitw: use `contains` from `util`

kanitw: This condition is surely poorly written as both cases and if and else contain the same statements.   This is hard to read.  Why don't you combine the predicate in a better way?

src/query/shorthand.ts
```diff
@@ -49,6 +50,59 @@ export function vlSpec(vlspec: ExtendedUnitSpec,
...
+    }, {} as Dict<boolean>);
+
+const CHANNEL_SUPPORTS_SORT: Dict<boolean> =
+  [Channel.X, Channel.Y, Channel.PATH, Channel.ORDER]
+    .reduce((m, channel) => {
+      m[channel] = true;
+      return m;
+    }, {} as Dict<boolean>);
+
+export const CHANNEL_SUPPORTS_PROPERTY =
```
kanitw: I'm sure there are smart way to make this chunk of code shorter.  Either 

PROPERTY_SUPPORTED_CHANNELS[prop][channel] 

channelSupportedProperties[prop][channel]

would work.  You just need to think more.

kanitw: I'm sure there are smart way to make this chunk of code shorter.  Either 

`PROPERTY_SUPPORTED_CHANNELS[prop][channel] `
or
`CHANNEL_SUPPORTED_PROPERTY[channel][prop]`

would work.  You just need to think more.

kanitw: Maybe I should just give you the answer.  But isn't this obvious?  I don't know why you need the complicated 2 layers of `reduce`.   

```
PROPERTY_SUPPORTED_CHANNELS = {
   sort: {x: true, ...}, 
   axis: {x: true, ...},
   ...
}
```

src/model.ts
```diff
@@ -694,7 +694,15 @@ export class SpecQueryModel {
 
         // otherwise, assign the proper to the field def
         if (encQ[prop] !== undefined) {
-          fieldDef[prop] = encQ[prop];
+
+          // axis, legend, scale, and sort only support certain channels
+          let propNeedsToCheckChannel: boolean = contains([Property.AXIS, Property.LEGEND, Property.SCALE, Property.SORT], prop);
+
+          // We only output axis, legend, scale, or sort that have correct supporting channel
+          if ((propNeedsToCheckChannel && !isEnumSpec(encQ.channel) && PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel]) ||
```
kanitw: Looks like you're excluding all these property is channel is an enumSpec.  
Add a test case for that and fix this!

Use 
```
if (isEnumSpec(encQ.channel) ||  // channel is enum spec, so the property might be supported 
  !PROPERTY_SUPPORTED_CHANNELS[prop] ||  // all channel support this prop
  PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel])
) {
  ...
}
```

and you can get rid of `propNeedsToCheckChannel` call that uses `contains`, which is unnecessary O(n).

test/model.test.ts
```diff
@@ -293,6 +293,62 @@ describe('SpecQueryModel', () => {
...
+        data: {values: [{A: 1}]},
+        transform: {filter: 'datum.A===1'},
+        mark: Mark.BAR,
+        encoding: {
+          color: {field: 'B', type: Type.QUANTITATIVE, legend: {orient: 'right', labelAlign: 'left', symbolSize: 12, title: 'test title'}}
+        },
+        config: DEFAULT_SPEC_CONFIG
+      });
+    });
+
```
kanitw: Missing channel = ? case

PR #238, opened 2016-08-17
Relax hasAllRequiredChannelsForMark constraint for when mark is an enum spec

### LINKED ISSUE #235: Relax constraint when there is only row/column
Currently, this query

```
{
  "spec": {
    "mark": "?",
    "encodings": [
      {
        "channel": "column",
        "field": "Horsepower",
        "type": "quantitative",
        "bin": {}
      }
    ],
    "config": {"facet": {"cell": {"width": 150,"height": 150}}}
  },
  "groupBy": ["field","aggregate","bin","timeUnit","stack","channel"],
  "chooseBy": "effectiveness",
  "config": {"omitTableWithOcclusion": false, "verbose": true}
}
```

returns in nothing due to `hasAllRequiredChannelsForMark`.   At least users to deserve to see something (although it's not good).

### LINKED ISSUE #237: Relax constraint when there is no x and y
```
{
  "spec": {
    "mark": "?",
    "encodings": [{"channel": "size","field": "Acceleration","type": "quantitative"}],
    "config": {"facet": {"cell": {"width": 150,"height": 150}}}
  },
  "groupBy": ["field","aggregate","bin","timeUnit","stack","channel"],
  "chooseBy": "effectiveness",
  "config": {"omitTableWithOcclusion": false}
}
```

This shouldn't return nothing.  (At least show a point plot.)

felixcodes:
fix #237 
- Allow a point plot to generate when mark is an enum spec.

fix #235 
- Relax spec when there is only row or column channel.

### DIFF
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -225,7 +225,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
           return specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
         case Mark.POINT:
           // This allows generating a point plot if channel was not an enum spec.
-          return !specM.enumSpecIndex.hasProperty(Property.CHANNEL) || 
+          return !specM.enumSpecIndex.hasProperty(Property.CHANNEL) ||
                  specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
       }
       /* istanbul ignore next */
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -224,7 +224,8 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
         case Mark.RULE:
           return specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
         case Mark.POINT:
-          return !specM.enumSpecIndex.hasProperty(Property.CHANNEL) || // This allows a point plot to generate if channel was enumspec
+          // This allows generating a point plot if channel was not an enum spec.
+          return !specM.enumSpecIndex.hasProperty(Property.CHANNEL) || 
                  specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
       }
       /* istanbul ignore next */
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -224,7 +224,7 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
         case Mark.RULE:
           return specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
         case Mark.POINT:
-          return !!specM.enumSpecIndex.mark || // This allows a point plot to generate if x or y aren't used but mark was ?
+          return !specM.enumSpecIndex.hasProperty(Property.CHANNEL) || // This allows a point plot to generate if channel was enumspec
                  specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
       }
       /* istanbul ignore next */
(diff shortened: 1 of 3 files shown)


src/constraint/spec.ts
```diff
@@ -219,11 +219,13 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
           return specM.channelUsed(Channel.TEXT);
         case Mark.BAR:
         case Mark.CIRCLE:
-        case Mark.POINT:
         case Mark.SQUARE:
         case Mark.TICK:
         case Mark.RULE:
           return specM.channelUsed(Channel.X) || specM.channelUsed(Channel.Y);
+        case Mark.POINT:
+          return isEnumSpec(specM.enumSpecIndex.mark) ? true :
```
kanitw: Do you need to call `isEnumSpec(specM.enumSpecIndex.mark)`  here?

Isn't `specM.enumSpecIndex.mark` sufficient?

kanitw: Plus `A ? true : B || C`  = `A || B || C`

kanitw: Why don't you do 
```
return A || // comment
   B || C; 
```

test/generate.test.ts
```diff
@@ -19,6 +19,46 @@ const CONFIG_WITH_AUTO_ADD_COUNT = extend({}, DEFAULT_QUERY_CONFIG, {autoAddCoun
...
+            channel: '?',
+            field: 'A',
+            type: Type.QUANTITATIVE
+          }]
+        };
+        const answerSet = generate(query, schema);
+        assert.equal(answerSet.length, 4);
+
+        assert.equal(answerSet[0].getMark(), Mark.POINT);
+        assert.equal(answerSet[0].getEncodingQueryByIndex(0).channel, Channel.X);
```
kanitw: Are you simply getting lucky with other constraint covers your constraint's bug by chance?  I don't understand why SIZE / SHAPE should be excluded given your logic.

test/generate.test.ts
```diff
@@ -19,6 +19,50 @@ const CONFIG_WITH_AUTO_ADD_COUNT = extend({}, DEFAULT_QUERY_CONFIG, {autoAddCoun
...
+        assert.equal(answerSet[1].getMark(), Mark.TICK);
+        assert.equal(answerSet[1].getEncodingQueryByIndex(0).channel, Channel.X);
+        assert.equal(answerSet[2].getMark(), Mark.POINT);
+        assert.equal(answerSet[2].getEncodingQueryByIndex(0).channel, Channel.Y);
+        assert.equal(answerSet[3].getMark(), Mark.TICK);
+        assert.equal(answerSet[3].getEncodingQueryByIndex(0).channel, Channel.Y);
+        assert.equal(answerSet[4].getEncodingQueryByIndex(0).channel, Channel.SIZE);
```
kanitw: Instead making the function correct, why do you make the test to be wrong instead?

src/constraint/spec.ts
```diff
@@ -378,6 +380,11 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     allowEnumSpecForProperties: false,
     strict: false,
     satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      if (specM.enumSpecIndex.hasProperty(Property.CHANNEL) &&
```
kanitw: I don't understand why you add this condition.  Looks like a hack.  
At least you should add comment justifying your design decision (if it's well principled, not a hack.)

test/generate.test.ts
```diff
@@ -19,6 +19,50 @@ const CONFIG_WITH_AUTO_ADD_COUNT = extend({}, DEFAULT_QUERY_CONFIG, {autoAddCoun
...
+            type: Type.QUANTITATIVE
+          }]
+        };
+        const answerSet = generate(query, schema);
+        assert.equal(answerSet.length, 1);
+        assert.equal(answerSet[0].getMark(), Mark.POINT);
+      });
+    });
+
+    describe('Q with mark=?, channel=?', () => {
```
kanitw: Add 'Q mark = point, channel = ?' 
and you should get only x y

PR #239, opened 2016-08-18
Add timeUnitHasVariation method for schema

RileyChang:
Add method `timeUnitHasVariation`. Returns true if the number of dates that are completely distinct in all "parts" of the timeUnit is greater than 1.

Part of a follow up to https://github.com/uwdata/voyager2/issues/93.

### DIFF
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -162,7 +162,7 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: false,
     satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
       if (encQ.timeUnit && encQ.type === Type.TEMPORAL) {
-        return schema.cardinality(encQ, false, true) > 1;
+        return schema.timeUnitHasVariation(encQ);
       }
       return true;
     }
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -155,7 +155,7 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
       return true;
     }
   },{
-    name: 'timeUnitShouldShowVariation',
+    name: 'timeUnitShouldHaveVariation',
     description: 'A particular time unit should be applied only if they produce unique values.',
     properties: [Property.TIMEUNIT, Property.TYPE],
     allowEnumSpecForProperties: false,
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -199,13 +199,18 @@ export class Schema {
   }
 
   /**
-   * Given an EncodingQuery with a timeUnit, returns true if the number of
-   * dates that are completely distinct in all parts of the timeUnit is greater than 1.
+   * Given an EncodingQuery with a timeUnit, returns true if the date field
+   * has multiple distinct values for all parts of the timeUnit. Returns undefined
+   * if the timeUnit is undefined.
    * i.e.
    * ('yearmonth', [Jan 1 2000, Feb 2 2000] returns false)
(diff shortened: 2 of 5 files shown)


kanitw:
I'm appending "for schema" to the title :)

src/constraint/encoding.ts
```diff
@@ -162,7 +162,7 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: false,
     satisfy: (encQ: EncodingQuery, schema: Schema, opt: QueryConfig) => {
       if (encQ.timeUnit && encQ.type === Type.TEMPORAL) {
-        return schema.cardinality(encQ, false, true) > 1;
+        return schema.timeUnitHasVariation(encQ);
```
kanitw: to make it consistent with the method, rename constraint to `timeUnitShouldHaveVariation`?

src/schema.ts
```diff
@@ -198,6 +198,35 @@ export class Schema {
...
+   * dates that are completely distinct in all parts of the timeUnit is greater than 1.
+   * i.e.
+   * ('yearmonth', [Jan 1 2000, Feb 2 2000] returns false)
+   * ('yearmonth', [Jan 1 2000, Feb 2 2001] returns true)
+   */
+  public timeUnitHasVariation(encQ: EncodingQuery): boolean {
+    // if there is no variation in `date`, there should not be variation in `day`
+    if (encQ.timeUnit === TimeUnit.DAY) {
+      const dateEncQ: EncodingQuery = extend({}, encQ);
+      dateEncQ.timeUnit = TimeUnit.DATE;
```
kanitw: you shouldn't modify a `const`  ... put `{timeUnit: ...}` as the last param in the `extend`?

src/schema.ts
```diff
@@ -198,6 +198,35 @@ export class Schema {
...
 
+  /**
+   * Given an EncodingQuery with a timeUnit, returns true if the number of
+   * dates that are completely distinct in all parts of the timeUnit is greater than 1.
+   * i.e.
+   * ('yearmonth', [Jan 1 2000, Feb 2 2000] returns false)
+   * ('yearmonth', [Jan 1 2000, Feb 2 2001] returns true)
+   */
+  public timeUnitHasVariation(encQ: EncodingQuery): boolean {
+    // if there is no variation in `date`, there should not be variation in `day`
```
kanitw: If timeUnit is undefined, unclear what would it return?  Should it return `undefined`?  Because it's unclear in that case? 

(And make your constraint only calls this if timeUnit is defined)

src/schema.ts
```diff
@@ -198,6 +198,35 @@ export class Schema {
     }
   }
 
+  /**
+   * Given an EncodingQuery with a timeUnit, returns true if the number of
```
kanitw: This reads quite confusing.  

> returns true if the number of
 dates that are completely distinct in all parts of the timeUnit is greater than 1.

Maybe

> returns true if the date field has multiple distinct values for all parts of the timeUnit?

src/schema.ts
```diff
@@ -198,6 +198,35 @@ export class Schema {
...
+      const dateEncQ: EncodingQuery = extend({}, encQ);
+      dateEncQ.timeUnit = TimeUnit.DATE;
+      if (this.cardinality(dateEncQ, false, true) <= 1) {
+        return false;
+      }
+    }
+
+    let fullTimeUnit = encQ.timeUnit;
+    for (let singleUnit of SINGLE_TIMEUNITS) {
+      if (containsTimeUnit(fullTimeUnit as TimeUnit, singleUnit)) {
```
kanitw: This is slightly inefficient as we have to always loop through all of them.  Maybe file an issue so that we can refactor all occurrence of `containsTimeUnit`?  (No need to do now.)

PR #244, opened 2016-08-19
Extend orderBy, chooseBy, orderGroupBy to support array

### LINKED ISSUE #195: Extend orderBy to support array (sort by a, then by b, ...)
Basically we need to extend `orderBy`, `chooseBy` to support array.

felixcodes:
Fix #195 
- Support string array type for orderBy, chooseBy, and orderGroupBy

### DIFF
--- a/src/ranking/ranking.ts
+++ b/src/ranking/ranking.ts
@@ -85,18 +85,38 @@ function getScore(model: SpecQueryModel, rankingName: string, schema: Schema, op
   return score;
 }
 
-export function comparator(name: string, schema: Schema, opt: QueryConfig) {
+export function comparator(name: string | string[], schema: Schema, opt: QueryConfig) {
   return (m1: SpecQueryModel, m2: SpecQueryModel) => {
-    return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
+    if (name instanceof Array) {
+      return getScoreDifference(name, m1, m2, schema, opt);
+
+    } else {
+      return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
+    }
   };
 }
 
-export function groupComparator(name: string, schema: Schema, opt: QueryConfig) {
+export function groupComparator(name: string | string[], schema: Schema, opt: QueryConfig) {
   return (g1: SpecQueryModelGroup, g2: SpecQueryModelGroup) => {
     const m1 = g1.getTopSpecQueryModel();
     const m2 = g2.getTopSpecQueryModel();
-    return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
+    if (name instanceof Array) {
+      return getScoreDifference(name, m1, m2, schema, opt);
+    } else {
+      return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
+    }
   };
+}
+
+function getScoreDifference(name: string[], m1: SpecQueryModel, m2: SpecQueryModel, schema, opt): number {
+    let scoreDifference = getScore(m2, name[0], schema, opt).score - getScore(m1, name[0], schema, opt).score;
+    for (let i = 1; i < name.length; i++) {
+      if (scoreDifference !== 0) {
+        break;
+      }
+      scoreDifference = getScore(m2, name[i], schema, opt).score - getScore(m1, name[i], schema, opt).score;
+    }
+    return scoreDifference;
 }
 
 export const EFFECTIVENESS = 'effectiveness';
--- a/src/ranking/ranking.ts
+++ b/src/ranking/ranking.ts
@@ -89,7 +89,6 @@ export function comparator(name: string | string[], schema: Schema, opt: QueryCo
   return (m1: SpecQueryModel, m2: SpecQueryModel) => {
     if (name instanceof Array) {
       return getScoreDifference(name, m1, m2, schema, opt);
-
     } else {
       return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
     }
--- a/src/ranking/ranking.ts
+++ b/src/ranking/ranking.ts
@@ -75,7 +75,7 @@ export function rank(group: SpecQueryModelGroup, query: Query, schema: Schema, l
   return group;
 }
 
-function getScore(model: SpecQueryModel, rankingName: string, schema: Schema, opt: QueryConfig) {
+export function getScore(model: SpecQueryModel, rankingName: string, schema: Schema, opt: QueryConfig) {
   if (model.getRankingScore(rankingName) !== undefined) {
     return model.getRankingScore(rankingName);
   }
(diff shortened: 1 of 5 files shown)


kanitw:
still missing test.

src/ranking/ranking.ts
```diff
@@ -85,20 +85,39 @@ function getScore(model: SpecQueryModel, rankingName: string, schema: Schema, op
...
+      return getScoreDifference(name, m1, m2, schema, opt);
+    } else {
+      return getScore(m2, name, schema, opt).score - getScore(m1, name, schema, opt).score;
+    }
   };
 }
 
+function getScoreDifference(name: string[], m1: SpecQueryModel, m2: SpecQueryModel, schema, opt): number {
+    let scoreDifference = getScore(m2, name[0], schema, opt).score - getScore(m1, name[0], schema, opt).score;
+    for (let i = 1; i < name.length; i++) {
```
kanitw: Why don't you start at 0 and just call `getScore` only once. 
Instead of break, you can just return things right away.  

And return 0 at the end.

test/query/query.test.ts
```diff
@@ -121,6 +122,34 @@ describe('query/query', () => {
       });
     });
 
+    describe('rank', () => {
+      const q: Query = {
```
kanitw: Why isn't this q a part of `it`.  It's not reused so you shouldn't pollute the outer scope.

test/ranking/ranking.test.ts
```diff
@@ -30,4 +34,56 @@ describe('ranking', () => {
...
+            bin: SHORT_ENUM_SPEC,
+            aggregate: SHORT_ENUM_SPEC,
+            field: 'Q1',
+            type: Type.QUANTITATIVE
+          }
+        ]
+      };
+      const answerSet = generate(specQ, schema, DEFAULT_QUERY_CONFIG);
+      const comparator = comparatorFactory(['aggregationQuality', 'effectiveness'], schema, DEFAULT_QUERY_CONFIG);
+      assert.isNumber(comparator(answerSet[1], answerSet[4]));
```
kanitw: This is a unit test for comparator factory.  It shouldn't call `generate`.  

Readers can't imagine what would be answerSet[1] and [4].  Plus, 1 and 4 are both arbitrary.  
Why don't you follow my suggestion that other day which is just create two SpecQueryModel.

test/ranking/ranking.test.ts
```diff
@@ -30,4 +34,56 @@ describe('ranking', () => {
...
+    });
+
+    it('should create a comparator that returns a score difference when passed an orderBy string', () => {
+      const specQ: SpecQuery = {
```
kanitw: Same issue

test/query/query.test.ts
```diff
@@ -121,6 +122,34 @@ describe('query/query', () => {
...
+
+        const output = query(q, schema);
+        const result = output.result;
+
+        for (let i = 1; i < result.items.length; i++) {
+          let previousItem = result.items[i-1];
+          let currentItem = result.items[i];
+
+          assert.isTrue(
+            getScore(previousItem as SpecQueryModel, 'aggregationQuality', schema, DEFAULT_QUERY_CONFIG).score >=
```
kanitw: Specifically, it should be

```
(aggregateQuality(left)  >  aggregateQuality(right)) || 
(aggregateQuality(left)  ===  aggregateQuality(right) && effectiveness(left)  >=  effectiveness(right))
```

test/ranking/ranking.test.ts
```diff
@@ -30,4 +31,64 @@ describe('ranking', () => {
...
+            {channel: Channel.X, field: 'N', type: Type.NOMINAL},
+            {channel: Channel.Y, field: 'Q', type: Type.QUANTITATIVE},
+          ]
+        },
+        schema,
+        DEFAULT_QUERY_CONFIG
+      );
+
+      const comparator = comparatorFactory(['aggregationQuality', 'effectiveness'], schema, DEFAULT_QUERY_CONFIG);
+      assert.isNumber(comparator(specM1, specM2));
```
kanitw: This doesn't tell if your comparator returns correct result!  

The test would pass even if you put a random number generator as the comparator function.

test/ranking/ranking.test.ts
```diff
@@ -30,4 +31,64 @@ describe('ranking', () => {
...
+            {channel: Channel.X, field: 'N', type: Type.NOMINAL},
+            {channel: Channel.Y, field: 'Q', type: Type.QUANTITATIVE},
+          ]
+        },
+        schema,
+        DEFAULT_QUERY_CONFIG
+      );
+
+      const comparator = comparatorFactory('aggregationQuality', schema, DEFAULT_QUERY_CONFIG);
+      assert.isNumber(comparator(specM1, specM2));
```
kanitw: This doesn't tell if your comparator returns correct result!  

The test would pass even if you put a random number generator as the comparator function.

test/ranking/ranking.test.ts
kanitw: This have two cases covered

> a) (aggregateQuality(left)  >  aggregateQuality(right)) 

Try  QxQ > MEAN(Q)xMEAN(Q)

> b) (aggregateQuality(left)  ===  aggregateQuality(right) && effectiveness(left)  >=  effectiveness(right))

Try something that should be plot with line then copy it and change to point.  Than Line > Point and the data is the same.  

You can reuse this example for the case that we input just `'aggregationQuality'` like below.

test/ranking/ranking.test.ts
```diff
@@ -33,13 +34,14 @@ describe('ranking', () => {
   });
 
   describe('comparatorFactory', () => {
-    it('should create a comparator that returns a score difference when passed an orderBy array', () => {
+    it.only('should create a comparator that returns a score difference when passed an orderBy array', () => {
```
kanitw: The result comparator should return positive value when ...

test/ranking/ranking.test.ts
```diff
@@ -48,21 +50,53 @@ describe('ranking', () => {
...
+            {channel: Channel.COLOR, field: 'symbol', type: Type.NOMINAL}
           ]
         },
         schema,
         DEFAULT_QUERY_CONFIG
       );
 
       const comparator = comparatorFactory(['aggregationQuality', 'effectiveness'], schema, DEFAULT_QUERY_CONFIG);
-      assert.isNumber(comparator(specM1, specM2));
+      assert.isTrue(comparator(specM1, specM2) > 0);
```
kanitw: I think there is `assert.lessThan` which will true better error if something is wrong.

Using `isTrue` for <, > would only produce something like `expect true to be false` when something goes wrong.

test/ranking/ranking.test.ts
```diff
@@ -30,4 +33,98 @@ describe('ranking', () => {
...
+            {aggregate: AggregateOp.MEAN, channel: Channel.X, field: 'Q', type: Type.QUANTITATIVE},
+            {aggregate: AggregateOp.MEAN, channel: Channel.Y, field: 'Q1', type: Type.QUANTITATIVE},
+          ]
+        },
+        schema,
+        DEFAULT_QUERY_CONFIG
+      );
+
+      const comparator = comparatorFactory('aggregationQuality', schema, DEFAULT_QUERY_CONFIG);
+      assert.isTrue(comparator(specM1, specM2) < 0);
```
kanitw: isTrue

PR #245, opened 2016-08-22
Add constraint: omitNonLinearScaleTypeWithStack

### LINKED ISSUE #240: Scale Type Constraint for Stack
Basically for stacked chart, we should only use linear scale.  All non-linear scales are bad!

felixcodes:
Fix #240 
- Add omitNonLinearScaleTypeWithStack constraint and test

### DIFF
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -537,15 +537,18 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     name: 'omitNonLinearScaleTypeWithStack',
     description: 'Stacked plot should only use linear scale',
     properties: [Property.CHANNEL, Property.MARK, Property.AGGREGATE, Property.AUTOCOUNT, Property.SCALE, Property.SCALE_TYPE, Property.TYPE],
+    // TODO: Property.STACK
     allowEnumSpecForProperties: false,
     strict: true,
     satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
       const stack = specM.stack();
       if (stack) {
         for (let encQ of specM.getEncodings()) {
-          if ((!!encQ.aggregate || encQ.autoCount === true) && (encQ.type === Type.QUANTITATIVE)) {
-            if (scaleType(encQ) !== ScaleType.LINEAR) {
-              return false;
+          if ((!!encQ.aggregate || encQ.autoCount === true) &&
+             encQ.type === Type.QUANTITATIVE &&
+             contains([Channel.X, Channel.Y], encQ.channel)) {
+              if (scaleType(encQ) !== ScaleType.LINEAR) {
+                return false;
             }
           }
         }
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -549,6 +549,26 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'omitStackWithNonLinearScaleType',
+    description: 'Stacked plot should only use linear scale',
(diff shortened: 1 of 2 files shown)


src/constraint/spec.ts
```diff
@@ -534,6 +534,26 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'omitNonLinearScaleTypeWithStack',
+    description: 'Stacked plot should only use linear scale',
+    properties: [Property.CHANNEL, Property.MARK, Property.AGGREGATE, Property.AUTOCOUNT, Property.SCALE, Property.SCALE_TYPE],
```
kanitw: Missing `property.type`

src/constraint/spec.ts
```diff
@@ -534,6 +534,26 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
     }
   },
   {
+    name: 'omitNonLinearScaleTypeWithStack',
+    description: 'Stacked plot should only use linear scale',
+    properties: [Property.CHANNEL, Property.MARK, Property.AGGREGATE, Property.AUTOCOUNT, Property.SCALE, Property.SCALE_TYPE, Property.TYPE],
```
kanitw: Add // TODO: Property.STACK 
to remind about future

src/constraint/spec.ts
```diff
@@ -534,6 +534,26 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+    allowEnumSpecForProperties: false,
+    strict: true,
+    satisfy: (specM: SpecQueryModel, schema: Schema, opt: QueryConfig) => {
+      const stack = specM.stack();
+      if (stack) {
+        for (let encQ of specM.getEncodings()) {
+          if ((!!encQ.aggregate || encQ.autoCount === true) && (encQ.type === Type.QUANTITATIVE)) {
```
kanitw: I think this is only for x and y? 

For example, if you group things by detail for the stack and map color to another aggregate field -- color can use non-linear scale.

(Add this as a test case)

test/constraint/spec.test.ts
```diff
@@ -765,6 +765,63 @@ describe('constraints/spec', () => {
...
+      assert.isTrue(SPEC_CONSTRAINT_INDEX['omitNonLinearScaleTypeWithStack'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
+    });
+
+    it('should return true if color uses a non-linear scale when it is mapped to a non-X or non-Y channel that is aggregate', () => {
+      const specM = buildSpecQueryModel({
+        mark: Mark.BAR,
+        encodings: [
+          {channel: Channel.X, field: 'A', type: Type.QUANTITATIVE, aggregate: AggregateOp.SUM},
+          {channel: Channel.Y, field: 'B', type: Type.NOMINAL},
+          {channel: Channel.COLUMN, field: 'C', type: Type.NOMINAL, aggregate: AggregateOp.ARGMAX},
```
kanitw: ARGMAX??? This doesn't sounds sensible.

kanitw: I don't even know why you need column here.  Seems random

test/constraint/spec.test.ts
```diff
@@ -765,6 +765,63 @@ describe('constraints/spec', () => {
...
+          {channel: Channel.COLOR, field: 'C', type: Type.NOMINAL, scale: {type: ScaleType.ORDINAL}},
```
kanitw: I don't see why this test case is making any difference from above since nominal always use ORDINAL scale already.  

What you actually should test is mapping quantitative field to color with non-linear scale (e.g., pow)

PR #249, opened 2016-08-25
Add getFieldSchemas method to Schema class

RileyChang:
Add a public accessor to the Schema class for retrieving a list of field definitions.

Part of https://github.com/uwdata/voyager2/issues/126


kanitw:
LGTM, feel free to merge after the check finishes running and pass.

PR #250, opened 2016-08-26
Shorthand Parser

### LINKED ISSUE #229: cql.query.shorthand.parse
- [x] Basic Version: We don't deal with `||` inside filter (so we can just split the shorthand by `|`)
- [ ] Advance Version:  Deal with that.  

Get a SpecQuery out of the shorthand

felixcodes:
fix #229

### DIFF
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -140,11 +140,9 @@ export function spec(specQ: SpecQuery,
 }
 
 export function calculate(formulaArr: Formula[]): string {
-  return formulaArr.map(function(calculateItem) {
-    // TODO(https://github.com/uwdata/compassql/issues/260)
-    // This should be in the form {field1:expr1,field2:expr2, ... ,fieldN:exprN}
-    return `{${calculateItem.field}:${calculateItem.expr}}`;
-  }).join(',');
+  return '{' + formulaArr.map(function(calculateItem) {
+    return `${calculateItem.field}:${calculateItem.expr}`;
+  }).join(',') + '}';
 }
 
 /**
@@ -309,18 +307,16 @@ export function parse(shorthand: string): SpecQuery {
 
     if (splitPartKey === 'calculate') {
       let transformQ: TransformQuery = specQ.transform || {};
-
       let calculate = [];
-      let formulas = splitPartValue.split(',');
+      let formulas = splitPartValue
+                      .substring(1, splitPartValue.length - 1)
+                      .split(',');
 
       for (let formulaString of formulas) {
         let formula: Formula = {} as Formula;
-        // FIXME(https://github.com/uwdata/compassql/issues/260)
-        // Calculate shorthand should be in the form of `calculate: {field1:expr1, field2: expr2}
-        // This parser will need to be updated accordingly in the future
         let formulaParts = formulaString.split(':');
-        formula.field = formulaParts[0].substr(1);
-        formula.expr = formulaParts[1].slice(0, -1);
+        formula.field = formulaParts[0];
+        formula.expr = formulaParts[1];
         calculate.push(formula);
       }
 
@@ -353,7 +349,7 @@ export function parse(shorthand: string): SpecQuery {
  * @param delim The delimiter string used to separate the string
  * @param number The value used to determine how many times the string is split
  */
-export function splitWithTail(str: string, delim: string, count: number) {
+export function splitWithTail(str: string, delim: string, count: number): string[] {
   let result = [];
   let lastIndex = 0;
 
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -302,22 +302,44 @@ export function parse(shorthand: string): SpecQuery {
     const splitPartValue = splitPart[1];
 
     if (CHANNEL_INDEX[splitPartKey] || splitPartKey === '?') {
-      shorthandParser.encoding(specQ, splitPartKey, splitPartValue);
+      const encQ = shorthandParser.encoding(splitPartKey, splitPartValue);
+      specQ.encodings.push(encQ);
       continue;
     }
 
     if (splitPartKey === 'calculate') {
-      shorthandParser.calculate(specQ, splitPartValue);
+      let transformQ: TransformQuery = specQ.transform || {};
+
+      let calculate = [];
+      let formulas = splitPartValue.split(',');
+
+      for (let formulaString of formulas) {
+        let formula: Formula = {} as Formula;
+        // FIXME(https://github.com/uwdata/compassql/issues/260)
+        // Calculate shorthand should be in the form of `calculate: {field1:expr1, field2: expr2}
+        // This parser will need to be updated accordingly in the future
+        let formulaParts = formulaString.split(':');
+        formula.field = formulaParts[0].substr(1);
+        formula.expr = formulaParts[1].slice(0, -1);
+        calculate.push(formula);
+      }
+
+      transformQ.calculate = calculate;
+      specQ.transform = transformQ;
       continue;
     }
 
     if (splitPartKey === 'filter') {
-      shorthandParser.filter(specQ, splitPartValue);
+      let transformQ: TransformQuery = specQ.transform || {};
+      transformQ.filter = JSON.parse(splitPartValue);
+      specQ.transform = transformQ;
       continue;
     }
 
     if (splitPartKey === 'filterInvalid') {
-      shorthandParser.filterInvalid(specQ, splitPartValue);
+      let transformQ: TransformQuery = specQ.transform || {};
+      transformQ.filterInvalid = JSON.parse(splitPartValue);
+      specQ.transform = transformQ;
       continue;
     }
   }
@@ -332,15 +354,35 @@ export function parse(shorthand: string): SpecQuery {
  * @param number The value used to determine how many times the string is split
  */
 export function splitWithTail(str: string, delim: string, count: number) {
-  let parts = str.split(delim);
-  let tail = parts.slice(count).join(delim);
-  let result = parts.slice(0, count);
-  result.push(tail);
+  let result = [];
+  let lastIndex = 0;
+
+  for (let i = 0; i < count; i++) {
+    let indexOfDelim = str.indexOf(delim, lastIndex);
+
+    if (indexOfDelim !== -1) {
+      result.push(str.substring(lastIndex, indexOfDelim));
+      lastIndex = indexOfDelim + 1;
+    } else {
+      break;
+    }
+  }
+
+  result.push(str.substr(lastIndex));
+
+  // If the specified count is greater than the number of delimiters that exist in the string,
+  // an empty string will be pushed count minus number of delimiter occurence times.
+  if (result.length !== count + 1) {
+    while (result.length !== count + 1) {
+      result.push('');
+    }
+  }
+
   return result;
 }
 
 export namespace shorthandParser {
-  export function encoding(specQ: SpecQuery, channel: string, fieldDefShorthand: string) {
+  export function encoding(channel: string, fieldDefShorthand: string): EncodingQuery {
     let encQ: EncodingQuery = {channel: channel};
 
     if (fieldDefShorthand.indexOf('(') !== -1) {
@@ -349,7 +391,7 @@ export namespace shorthandParser {
       encQ = rawFieldDef(encQ, splitWithTail(fieldDefShorthand, ',', 2));
     }
 
-    specQ.encodings.push(encQ);
+    return encQ;
   }
 
   export function rawFieldDef(encQ: EncodingQuery, fieldDefPart: string[]): EncodingQuery {
@@ -383,6 +425,7 @@ export namespace shorthandParser {
             )
           );
 
+          // TODO(https://github.com/uwdata/compassql/issues/97): Make this generalized for other bin properties.
           if (nestedPropertyParent === 'maxbins') {
             encQ.bin['maxbins'] = parsedValue;
           } else {
@@ -415,13 +458,13 @@ export namespace shorthandParser {
       for (let encodingProperty in fnEnumIndex) {
         if (isArray(fnEnumIndex[encodingProperty])) {
           encQ[encodingProperty] = {enum: fnEnumIndex[encodingProperty]};
-        } else {
+        } else { // Definitely a `SHORT_ENUM_SPEC`
           encQ[encodingProperty] = fnEnumIndex[encodingProperty];
         }
       }
 
       return rawFieldDef(encQ,
-        splitWithTail(fieldDefShorthand.substring(closingBraceIndex + 2, fieldDefShorthand.length), ',', 2)
+        splitWithTail(fieldDefShorthand.substring(closingBraceIndex + 2, fieldDefShorthand.length - 1), ',', 2)
       );
     } else {
       let func = fieldDefShorthand.substring(0, fieldDefShorthand.indexOf('('));
@@ -429,58 +472,15 @@ export namespace shorthandParser {
       let insideFnParts = splitWithTail(insideFn, ',', 2);
 
       if (AGGREGATE_OP_INDEX[func]) {
-        return aggregate(encQ, func, insideFnParts);
+        encQ.aggregate = func;
+        return rawFieldDef(encQ, insideFnParts);
       } else if (MULTI_TIMEUNIT_INDEX[func] || SINGLE_TIMEUNIT_INDEX[func]) {
-        return timeUnit(encQ, func, insideFnParts);
+        encQ.timeUnit = func;
+        return rawFieldDef(encQ, insideFnParts);
       } else if (func === 'bin') {
-        return bin(encQ, func, insideFnParts);
+        encQ.bin = {};
+        return rawFieldDef(encQ, insideFnParts);
       }
     }
-  }
-
-  export function aggregate(encQ: EncodingQuery, aggregate: string, insideFnParts: string[]): EncodingQuery {
-    encQ.aggregate = aggregate;
-    return rawFieldDef(encQ, insideFnParts);
-  }
-
-  export function timeUnit(encQ: EncodingQuery, timeUnit: string, insideFnParts: string[]): EncodingQuery {
-    encQ.timeUnit = timeUnit;
-    return rawFieldDef(encQ, insideFnParts);
-  }
-
-  export function bin(encQ: EncodingQuery, bin: string, insideFnParts: string[]): EncodingQuery {
-    encQ.bin = {};
-    return rawFieldDef(encQ, insideFnParts);
-  }
-
-  export function calculate(specQ: SpecQuery, splitPartValue: string) {
-    let transformQ: TransformQuery = specQ.transform || {};
-
-    let calculate = [];
-    let formulas = splitPartValue.split(',');
-
-    for (let formulaString of formulas) {
-      let formula: Formula = {} as Formula;
-      // FIXME(https://github.com/uwdata/compassql/issues/260)
-      let formulaParts = formulaString.split(':');
-      formula.field = formulaParts[0].substr(1);
-      formula.expr = formulaParts[1].slice(0, -1);
-      calculate.push(formula);
-    }
-
-    transformQ.calculate = calculate;
-    specQ.transform = transformQ;
-  }
-
-  export function filter(specQ: SpecQuery, splitPartValue: string) {
-    let transformQ: TransformQuery = specQ.transform || {};
-    transformQ.filter = JSON.parse(splitPartValue);
-    specQ.transform = transformQ;
-  }
-
-  export function filterInvalid(specQ: SpecQuery, invalid: string) {
-    let transformQ: TransformQuery = specQ.transform || {};
-    transformQ.filterInvalid = JSON.parse(invalid);
-    specQ.transform = transformQ;
   }
 }
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -330,16 +330,13 @@ export function parse(shorthand: string): SpecQuery {
 
     if (splitPartKey === 'calculate') {
       let transformQ: TransformQuery = specQ.transform || {};
-      let calculate = [];
-      let formulas = splitPartValue
-                      .substring(1, splitPartValue.length - 1)
-                      .split(',');
+      let calculate: Formula[] = [];
+      let fieldAndExpr = JSON.parse(splitPartValue);
 
-      for (let formulaString of formulas) {
+      for (let field in fieldAndExpr) {
         let formula: Formula = {} as Formula;
-        let formulaParts = formulaString.split(':');
-        formula.field = formulaParts[0];
-        formula.expr = formulaParts[1];
+        formula['field'] = field;
+        formula['expr'] = fieldAndExpr[field];
         calculate.push(formula);
       }
 
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -12,7 +12,7 @@ import {TransformQuery} from './transform';
 
 import {isEnumSpec, SHORT_ENUM_SPEC} from '../enumspec';
 import {getNestedEncodingPropertyChildren, Property, DEFAULT_PROPERTY_PRECEDENCE} from '../property';
-import {Dict, extend, keys} from '../util';
+import {Dict, extend, keys, isArray} from '../util';
 
 export type Replacer = (s: string) => string;
 
@@ -409,12 +409,15 @@ export namespace shorthandParser {
     // Aggregate, Bin, TimeUnit as enum spec case
     if (fieldDefShorthand[0] === '?') {
       let closingBraceIndex = getClosingBraceIndex(1, fieldDefShorthand);
-      let enumFnString = fieldDefShorthand.substring(2, closingBraceIndex);
-      let enumFnValuePairs = enumFnString.split(',');
 
-      for (let enumFnValuePair of enumFnValuePairs) {
-        let fnValue = enumFnValuePair.split(':');
-        encQ[JSON.parse(fnValue[0])] = JSON.parse(fnValue[1]);
+      let fnEnumIndex = JSON.parse(fieldDefShorthand.substring(1, closingBraceIndex + 1));
+
+      for (let encodingProperty in fnEnumIndex) {
+        if (isArray(fnEnumIndex[encodingProperty])) {
+          encQ[encodingProperty] = {enum: fnEnumIndex[encodingProperty]};
+        } else {
+          encQ[encodingProperty] = fnEnumIndex[encodingProperty];
+        }
       }
 
       return rawFieldDef(encQ,
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -8,7 +8,6 @@ import {toMap, isString} from 'datalib/src/util';
 
 import {EncodingQuery} from './encoding';
 import {SpecQuery, stack, fromSpec} from './spec';
-import {TransformQuery} from './transform';
 
 import {isEnumSpec, SHORT_ENUM_SPEC} from '../enumspec';
 import {getNestedEncodingPropertyChildren, Property, DEFAULT_PROPERTY_PRECEDENCE} from '../property';
@@ -329,33 +328,27 @@ export function parse(shorthand: string): SpecQuery {
     }
 
     if (splitPartKey === 'calculate') {
-      let transformQ: TransformQuery = specQ.transform || {};
+      specQ.transform = specQ.transform || {};
       let calculate: Formula[] = [];
-      let fieldAndExpr = JSON.parse(splitPartValue);
+      let fieldExprMapping = JSON.parse(splitPartValue);
 
-      for (let field in fieldAndExpr) {
-        let formula: Formula = {} as Formula;
-        formula['field'] = field;
-        formula['expr'] = fieldAndExpr[field];
-        calculate.push(formula);
+      for (let field in fieldExprMapping) {
+        calculate.push({field: field, expr: fieldExprMapping[field]});
       }
 
-      transformQ.calculate = calculate;
-      specQ.transform = transformQ;
+      specQ.transform.calculate = calculate;
       continue;
     }
 
     if (splitPartKey === 'filter') {
-      let transformQ: TransformQuery = specQ.transform || {};
-      transformQ.filter = JSON.parse(splitPartValue);
-      specQ.transform = transformQ;
+      specQ.transform = specQ.transform || {};
+      specQ.transform.filter = JSON.parse(splitPartValue);
       continue;
     }
 
     if (splitPartKey === 'filterInvalid') {
-      let transformQ: TransformQuery = specQ.transform || {};
-      transformQ.filterInvalid = JSON.parse(splitPartValue);
-      specQ.transform = transformQ;
+      specQ.transform = specQ.transform || {};
+      specQ.transform.filterInvalid = JSON.parse(splitPartValue);
       continue;
     }
   }
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -140,9 +140,12 @@ export function spec(specQ: SpecQuery,
 }
 
 export function calculate(formulaArr: Formula[]): string {
-  return '{' + formulaArr.map(function(calculateItem) {
-    return `${calculateItem.field}:${calculateItem.expr}`;
-  }).join(',') + '}';
+  return JSON.stringify(
+    formulaArr.reduce((m, calculateItem) => {
+      m[calculateItem.field] = calculateItem.expr;
+      return m;
+    }, {})
+  );
 }
 
 /**
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -1,13 +1,16 @@
-import {Channel} from 'vega-lite/src/channel';
+import {AGGREGATE_OPS} from 'vega-lite/src/aggregate';
+import {Channel, CHANNELS} from 'vega-lite/src/channel';
 import {Formula} from 'vega-lite/src/transform';
 import {ExtendedUnitSpec} from 'vega-lite/src/spec';
-import {Type} from 'vega-lite/src/type';
-import {isString} from 'datalib/src/util';
+import {SINGLE_TIMEUNITS, MULTI_TIMEUNITS} from 'vega-lite/src/timeunit';
+import {Type, TYPE_FROM_SHORT_TYPE} from 'vega-lite/src/type';
+import {toMap, isString} from 'datalib/src/util';
 
 import {EncodingQuery} from './encoding';
 import {SpecQuery, stack, fromSpec} from './spec';
-import {isEnumSpec, SHORT_ENUM_SPEC} from '../enumspec';
+import {TransformQuery} from './transform';
 
+import {isEnumSpec, SHORT_ENUM_SPEC} from '../enumspec';
 import {getNestedEncodingPropertyChildren, Property, DEFAULT_PROPERTY_PRECEDENCE} from '../property';
 import {Dict, extend, keys} from '../util';
 
@@ -138,6 +141,8 @@ export function spec(specQ: SpecQuery,
 
 export function calculate(formulaArr: Formula[]): string {
   return formulaArr.map(function(calculateItem) {
+    // TODO(https://github.com/uwdata/compassql/issues/260)
+    // This should be in the form {field1:expr1,field2:expr2, ... ,fieldN:exprN}
     return `{${calculateItem.field}:${calculateItem.expr}}`;
   }).join(',');
 }
@@ -165,7 +170,7 @@ export function encoding(encQ: EncodingQuery,
 }
 
 /**
- * Returns a field definiton shorthand for an encoding query
+ * Returns a field definition shorthand for an encoding query
  * @param encQ an encoding query
  * @param include Dict Set listing property types (key) to be included in the shorthand
  * @param replace Dictionary of replace function for values of a particular property type (key)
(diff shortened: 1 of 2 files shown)


src/query/shorthand.ts
```diff
@@ -171,7 +174,7 @@ export function encoding(encQ: EncodingQuery,
 }
 
 /**
- * Returns a field definiton shorthand for an encoding query
+ * Returns a field definisortton shorthand for an encoding query
```
kanitw: typo..

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+  let specQ: SpecQuery = null;
+
+  // Mark
+  specQ.mark = splitShorthand[0];
+
+  for (let part of splitShorthand) {
+
+    // Encodings
+    for (let channel of [Channel.COLOR, Channel.COLUMN, Channel.DETAIL, Channel.LABEL, Channel.OPACITY,
```
kanitw: Don't we have array for all channels in VL?

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+      // Check if TimeUnit
+
+      // Check if Bin
+    }
+
+    return encodingAfterChannel(encQ, afterChannel);
+  }
+
+  export function encodingAfterChannel(encQ: EncodingQuery, afterChannel: string): EncodingQuery {
```
kanitw: fieldDef

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+    // encQ.channel = channel;
+
+    let afterChannel = part.split(':')[1];
+
+    // Aggregate, AutoCount, TimeUnit, Bin
+    if (afterChannel.indexOf('(') !== -1) {
+
+      // Check if Aggregate
+      // reorder so that most comment AggregateOp appear first
+      for (let aggregateOp of [AggregateOp.ARGMAX, AggregateOp.ARGMIN, AggregateOp.AVERAGE, AggregateOp.DISTINCT,
```
kanitw: Don't we have aggregate array in VL?

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+                         Channel.TEXT, Channel.X, Channel.X2, Channel.Y, Channel.Y2]) {
+      const prefix = channel + ':';
+      if (part.substr(0, prefix.length) === prefix) {
+        specQ.encodings.push(shorthandParser[channel](part));
+        break;
+      }
+    }
+
+    // Transform
+    for (let transformPart of ['calculate', 'filter', 'filterInvalid']) {
```
kanitw: If something is already a part of the first one, you shouldn't have to loop through here.

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+    return encQ;
+  }
+
+  export function aggregate(encQ: EncodingQuery, afterChannel): EncodingQuery {
+    let aggregateIndex = afterChannel.indexOf('(');
+    encQ.aggregate = afterChannel.substr(0, aggregateIndex);
+
+    // Redo this so it is passed to a method that handles the contents inside of sum(    )
+    // ex. a method that will parse q,q,scale={ }   etc..
+    encQ = encodingAfterChannel(encQ, afterChannel.substr(aggregateIndex + 1).slice(0, -1));
```
kanitw: redundant

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+    // ex:
+    //  channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {zero: true}
+    //  a,q,scale={"zero":true}
+    return encQ;
+  }
+
+  export function aggregate(encQ: EncodingQuery, afterChannel): EncodingQuery {
+    let aggregateIndex = afterChannel.indexOf('(');
+    encQ.aggregate = afterChannel.substr(0, aggregateIndex);
```
kanitw: extract function name in front of (

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+      }
+    }
+
+    // ex:
+    //  channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {zero: true}
+    //  a,q,scale={"zero":true}
+    return encQ;
+  }
+
+  export function aggregate(encQ: EncodingQuery, afterChannel): EncodingQuery {
```
kanitw: rename to `fn` and take property type in.

src/query/shorthand.ts
```diff
@@ -261,3 +264,164 @@ export function fieldDef(encQ: EncodingQuery,
...
+      const prefix = channel + ':';
+      if (part.substr(0, prefix.length) === prefix) {
+        specQ.encodings.push(shorthandParser[channel](part));
+        break;
+      }
+    }
+
+    // Transform
+    for (let transformPart of ['calculate', 'filter', 'filterInvalid']) {
+      const transformPrefix = transformPart + ':';
```
kanitw: this prefix logic *probably* can be inside the transform method.

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+      let func = afterChannel.substring(0, afterChannel.indexOf('('));
+      encQ = fn(encQ, func, afterChannel);
+    } else {
+      encQ = encodingAfterFieldDef(encQ, splitWithTail(afterChannel, ',', 2));
+    }
+
+    specQ.encodings.push(encQ);
+  }
+
+  export function encodingAfterFieldDef(encQ: EncodingQuery, fieldDefPart: string[]): EncodingQuery {
```
kanitw: fieldDef

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+    if (splitPartKey === 'filterInvalid') {
+      shorthandParser.filterInvalid(specQ, splitPartValue);
+      continue;
+    }
+  }
+
+  return specQ;
+}
+
+export function splitWithTail(str: string, delim: string, count: number) {
```
kanitw: Add

```
/**
 * 
 */
```

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+  let parts = str.split(delim);
+  let tail = parts.slice(count).join(delim);
+  let result = parts.slice(0, count);
+  result.push(tail);
+  return result;
+}
+
+export namespace shorthandParser {
+  export function encoding(specQ: SpecQuery, channel: string, afterChannel: string) {
```
kanitw: afterChannel => fieldDefShorthand?

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+    let encQ: EncodingQuery = {channel: channel};
+
+    if (afterChannel.indexOf('(') !== -1) {
```
kanitw: Shouldn't logic for `(` be in `fieldDef` ?

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+      if (afterFieldAndType[i] === '}') {
+        return i;
+      }
+    }
+  }
+
+  export function fn(encQ: EncodingQuery, func: string, afterChannel: string): EncodingQuery {
+    let insideFn = afterChannel.substring(func.length + 1, afterChannel.length - 1);
+    let insideFnParts = splitWithTail(insideFn, ',', 2);
+
```
kanitw: //TODO: missing ?

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+    encQ.type = TYPE_FROM_SHORT_TYPE[fieldDefPart[1].toUpperCase()];
+
+    let partParams = fieldDefPart[2];
+    let closingBraceIndex = 0;
+
+    // TODO: don't use a loop
+    for (let nestedParentProperty of ['maxbins', 'scale', 'sort', 'axis', 'legend']) {
+      let nestedParentPropertyIndex = partParams.indexOf(nestedParentProperty, closingBraceIndex);
+      if (nestedParentPropertyIndex !== -1) {
+        let characterAfterEquals = partParams[nestedParentPropertyIndex + nestedParentProperty.length + 1];
```
kanitw: You don't really need to declare this var

src/query/shorthand.ts
kanitw: maybe rawFieldDef because you actually don't include function in this one.

src/query/shorthand.ts
```diff
@@ -261,3 +264,206 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+
+    // TODO
+    // when writing filter (for now) assume you're using an array of filter objects
+  }
+
+  export function filterInvalid(specQ: SpecQuery, invalid: string) {
+    let transformQ: TransformQuery = specQ.transform || {};
+
+    if (invalid === 'true') {
```
kanitw: parse

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+const SINGLE_TIMEUNIT_INDEX = toMap(SINGLE_TIMEUNITS);
+const MULTI_TIMEUNIT_INDEX = toMap(MULTI_TIMEUNITS);
+
+export function parse(shorthand: string): SpecQuery {
+  let splitShorthand = shorthand.split('|');
+
+  let specQ: SpecQuery = {mark: undefined, encodings: [] as EncodingQuery[]};
+
+  // Mark
+  specQ.mark = splitShorthand[0];
```
kanitw: Why not replace undefined above with this `splitShorthand[0]`?

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+  for (let part of splitShorthand) {
+    const splitPart = splitWithTail(part, ':', 1);
```
kanitw: splittedPart

kanitw: Nvm, just realize 3rd form is also split.  ... english....

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+  }
+
+  export function rawFieldDef(encQ: EncodingQuery, fieldDefPart: string[]): EncodingQuery {
+    encQ.field = fieldDefPart[0];
+    encQ.type = TYPE_FROM_SHORT_TYPE[fieldDefPart[1].toUpperCase()];
+
+    let partParams = fieldDefPart[2];
+    let closingBraceIndex = 0;
+
+    // TODO: don't use a loop
```
kanitw: File an issue that you shouldn't use a loop here.  (But we can fix it later)

kanitw: Change TODO to FIXME.   (yes, this is not very good and i'd not let this slip if it's not a deadline.)

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+    for (let nestedParentProperty of ['maxbins', 'scale', 'sort', 'axis', 'legend']) {
```
kanitw: `nestedParentProperty` => `nestedPropertyParent`

+ Add another comment line

```
// FIXME: separate maxbins from nestedPropertyParent and support other bin properties. 
```

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+            encQ[nestedParentProperty] = parsedValue;
+          }
+        }
+      }
+    }
+
+    return encQ;
+  }
+
+  export function getClosingBraceIndex(openingBraceIndex: number, str: string): number {
```
kanitw: Extend this to support nested braces and brackets

kanitw: Add TODO

```
// TODO(https://github.com/uwdata/compassql/issues/259): 
// Extend this to support nested braces and brackets
```

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
   return (fn ? fn + '(' + fieldAndParams + ')' : fieldAndParams);
 }
+
+const CHANNEL_INDEX = toMap(CHANNELS);
+const AGGREGATE_OP_INDEX = toMap(AGGREGATE_OPS);
+const SINGLE_TIMEUNIT_INDEX = toMap(SINGLE_TIMEUNITS);
+const MULTI_TIMEUNIT_INDEX = toMap(MULTI_TIMEUNITS);
+
+export function parse(shorthand: string): SpecQuery {
+  let splitShorthand = shorthand.split('|');
```
kanitw:
```
// TODO(https://github.com/uwdata/compassql/issues/259): 
// Do not split directly, but use and upgraded version of `getClosingBraceIndex()`
```

src/query/shorthand.ts
```diff
@@ -255,3 +258,184 @@ export function fieldDef(encQ: EncodingQuery,
...
+      }
+    }
+  }
+
+  export function fn(encQ: EncodingQuery, fieldDefShorthand: string): EncodingQuery {
+    let func = fieldDefShorthand.substring(0, fieldDefShorthand.indexOf('('));
+    let insideFn = fieldDefShorthand.substring(func.length + 1, fieldDefShorthand.length - 1);
+    let insideFnParts = splitWithTail(insideFn, ',', 2);
+
+    // TODO: add support for ?
```
kanitw: file an issue and add issue url behind TODO

```
// TODO(<issue url>)
```

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,124 @@ describe('query/shorthand', () => {
...
+    describe('fn', () => {
+      it('should correctly parse an encoding query given a fieldDefShorthand with aggregation function', () => {
+        let encQ: EncodingQuery = {} as EncodingQuery;
+        shorthandParser.fn(encQ, 'sum(horsepower,q)');
+        assert.deepEqual(encQ, {aggregate: AggregateOp.SUM, field: 'horsepower', type: Type.QUANTITATIVE});
+      });
+
+      it('should correctly parse an encoding query given a fieldDefShorthand with count function', () => {
+        let encQ: EncodingQuery = {} as EncodingQuery;
+        shorthandParser.fn(encQ, 'count(model,n)');
```
kanitw: count model and nominal  This spec doesn't make sense at all.

count(*,q) --> this is how it works.

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,124 @@ describe('query/shorthand', () => {
...
+        assert.deepEqual(encQ,{aggregate: AggregateOp.COUNT, field: 'model', type: Type.NOMINAL});
+      });
+
+      it('should correctly parse an encoding query given a fieldDefShorthand with timeunit function', () => {
+        let encQ: EncodingQuery = {} as EncodingQuery;
+        shorthandParser.fn(encQ, 'hours(a,q)');
```
kanitw: why is hour quantitative?

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,124 @@ describe('query/shorthand', () => {
...
+      });
+
+      it('should correctly parse an encoding query given a fieldDefShorthand with bin function', () => {
+        let encQ: EncodingQuery = {} as EncodingQuery;
+        shorthandParser.fn(encQ, 'bin(a,q,maxbins=20,sort={"field":"a","op":"mean","order":"descending"})');
+        assert.deepEqual(encQ, {
+          bin: {maxbins: 20},
+          field: 'a',
+          type: Type.QUANTITATIVE,
+          sort: {field: 'a', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}
```
kanitw: You can't really sort bin with SortField.  You should create a separate test that sort either N or O field with SortField.

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,124 @@ describe('query/shorthand', () => {
...
+        });
+      });
+    });
+
+    describe('rawFieldDef', () => {
+      it('should correctly parse an encoding query from fieldDef parts', () => {
+        const encQ = shorthandParser.rawFieldDef({} as EncodingQuery,
+        splitWithTail('a,q,scale={"domain":["cats","dogs"],"exponent":3,"type":"pow"},axis={"orient":"top"},legend=false', ',', 2));
```
kanitw: You never have legend together with legend for the same field.

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,124 @@ describe('query/shorthand', () => {
...
+        assert.deepEqual(encQ, {axis: {orient: AxisOrient.TOP}, field: 'a', legend: false, scale: {domain: ['cats', 'dogs'], exponent: 3, type: ScaleType.POW}, type: Type.QUANTITATIVE});
+      });
+    });
+
+    describe('calculate', () => {
+      it('should correctly parse calculate from shorthand', () => {
+        let specQ: SpecQuery = {} as SpecQuery;
+        let shorthandPart = 'calculate:{x2:datum.x*2},{x3:datum.x*3}';
```
kanitw: Why is it `calculate:{x2:datum.x*2},{x3:datum.x*3}` ????????????????????

It should be
```
calculate:{x2:datum.x*2, x3:datum.x*3}
```

If the old code is wrong, please file an issue.  Luckily we don't use calculate in our user study.

src/query/shorthand.ts
```diff
@@ -255,3 +260,192 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+  export function calculate(specQ: SpecQuery, splitPartValue: string) {
+    let transformQ: TransformQuery = specQ.transform || {};
+
+    let calculate = [];
+    let formulas = splitPartValue.split(',');
+
+    for (let formulaString of formulas) {
+      let formula: Formula = {} as Formula;
+      let formulaParts = formulaString.split(':');
```
kanitw: Missing FIXME

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,144 @@ describe('query/shorthand', () => {
     });
   });
 
+  describe('parse', () => {
+    it('should correctly parse a shorthand string', () => {
+      let specQ = parse(
+        'point|calculate:{b2:3*datum["b2"]}|filter:"datum[\\"b2\\"] > 60"|filterInvalid:false|x:b2,q|y:bin(balance,q,sort={"field":"balance","op":"mean","order":"descending"})'
```
kanitw: bin here is still sorted....

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,144 @@ describe('query/shorthand', () => {
...
+        mark: Mark.POINT,
+        encodings: [
+          {channel: Channel.X, field: 'b2', type: Type.QUANTITATIVE},
+          {bin: {}, channel: Channel.Y, field: 'balance', type: Type.QUANTITATIVE, sort: {field: 'balance', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}}
+        ]
+      });
+    });
+
+    it('should correctly parse an ambiguous shorthand', () => {
+      let specQ = parse('?|?:?(?,?)');
```
kanitw: what about Q and T?

Q shouldn't have timeUnit (file an issue if u haven't handle this)

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,144 @@ describe('query/shorthand', () => {
...
+
+  describe('shorthandParser', () => {
+    describe('encoding', () => {
+      it('should correctly parse an encoding query given a channel and fieldDefShorthand', () => {
+        let specQ: SpecQuery = {mark: Mark.POINT, encodings: []} as SpecQuery;
+
+        shorthandParser.encoding(
+          specQ,
+          'x',
+          'bin(a,q,maxbins=20,scale={"type":"log"})'
```
kanitw: bin with log scale? meh

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,144 @@ describe('query/shorthand', () => {
...
+          field: 'a',
+          type: Type.QUANTITATIVE,
+        });
+      });
+    });
+
+    describe('rawFieldDef', () => {
+      it('should correctly parse an encoding query from fieldDef parts', () => {
+        let encQ = shorthandParser.rawFieldDef({} as EncodingQuery,
+          splitWithTail('a,q,scale={"domain":["cats","dogs"],"exponent":3,"type":"pow"},axis={"orient":"top"}', ',', 2)
```
kanitw: cats and dogs are not Q.... and they do not work with pow scale......

test/query/shorthand.test.ts
kanitw: Actually never mind ... my updated function in #262 would require rewriting this part anyway.

src/query/shorthand.ts
```diff
@@ -403,23 +403,32 @@ export namespace shorthandParser {
...
-    } else if (AGGREGATE_OP_INDEX[func]) {
-      return aggregate(encQ, func, insideFnParts);
-    } else if (MULTI_TIMEUNIT_INDEX[func] || SINGLE_TIMEUNIT_INDEX[func]) {
-      return timeUnit(encQ, func, insideFnParts);
-    } else if (func === 'bin') {
-      return bin(encQ, func, insideFnParts);
+    // Aggregate, Bin, TimeUnit as enum spec case
+    if (fieldDefShorthand[0] === '?') {
+      let closingBraceIndex = getClosingBraceIndex(1, fieldDefShorthand);
+      let enumFnString = fieldDefShorthand.substring(2, closingBraceIndex);
```
kanitw: This is definitely incorrect for
```
{"aggregate": ["max", "min"], "bin": [false,true]}
```

Why don't you just parse the JSON and check if it's '?' or array

If it's array, assign `{enum:  <array>}` otherwise, `SHORT_ENUM_SPEC`.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+}
+
+/**
+ * Split a string n times into substrings with the specified delimiter and return them as an array.
+ * @param str The string to be split
+ * @param delim The delimiter string used to separate the string
+ * @param number The value used to determine how many times the string is split
+ */
+export function splitWithTail(str: string, delim: string, count: number) {
+  let parts = str.split(delim);
```
kanitw: I think this is unnecessary expensive, especially when you use it with fieldDef with params. 
(Imagine how many `,` could there be inside your stringified JSON. 

Why don't you just call `indexOf` with correct startIndex for `count` times, each time extract the correct substring.  Then extract the last portion of substring at the end.  That will be way more efficient.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+          }
+
+          let parsedValue = JSON.parse(
+            partParams.substring(
+              nestedPropertyParentIndex + nestedPropertyParent.length + 1,
+              nextCommaIndex
+            )
+          );
+
+          if (nestedPropertyParent === 'maxbins') {
```
kanitw: Missing
```
// TODO(https://github.com/uwdata/compassql/issues/97): Make this generalize for other bin properties. 
```

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+    // Aggregate, Bin, TimeUnit as enum spec case
+    if (fieldDefShorthand[0] === '?') {
+      let closingBraceIndex = getClosingBraceIndex(1, fieldDefShorthand);
+
+      let fnEnumIndex = JSON.parse(fieldDefShorthand.substring(1, closingBraceIndex + 1));
+
+      for (let encodingProperty in fnEnumIndex) {
+        if (isArray(fnEnumIndex[encodingProperty])) {
+          encQ[encodingProperty] = {enum: fnEnumIndex[encodingProperty]};
+        } else {
```
kanitw:
```
else { // Definitely a `SHORT_ENUM_SPEC`
```

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+          encQ[encodingProperty] = fnEnumIndex[encodingProperty];
+        }
+      }
+
+      return rawFieldDef(encQ,
+        splitWithTail(fieldDefShorthand.substring(closingBraceIndex + 2, fieldDefShorthand.length), ',', 2)
```
kanitw: why `fieldDefShorthand.length`, not `fieldDefShorthand.length-1`?  

Don't you have to exclude `)` at the end?   (If the answer is you handle it inside `rawFieldDef` then it's wrong because rawFieldDef shouldn't assume knowledge of an outsider method.  This would probably be a kind of leaky abstraction.

kanitw: Plus, the method below does `fieldDefShorthand.length - 1`   so it's definitely inconsistent here.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+        return aggregate(encQ, func, insideFnParts);
+      } else if (MULTI_TIMEUNIT_INDEX[func] || SINGLE_TIMEUNIT_INDEX[func]) {
+        return timeUnit(encQ, func, insideFnParts);
+      } else if (func === 'bin') {
+        return bin(encQ, func, insideFnParts);
+      }
+    }
+  }
+
+  export function aggregate(encQ: EncodingQuery, aggregate: string, insideFnParts: string[]): EncodingQuery {
```
kanitw: aggregate, timeUnit, bin  all seem unnecessary.  They are all just one liners that do not worth extraction.  Why don't you just integrate them into `fn` method?  I don't why you would ever want to call these method separately.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+
+  export function calculate(specQ: SpecQuery, splitPartValue: string) {
+    let transformQ: TransformQuery = specQ.transform || {};
+
+    let calculate = [];
+    let formulas = splitPartValue.split(',');
+
+    for (let formulaString of formulas) {
+      let formula: Formula = {} as Formula;
+      // FIXME(https://github.com/uwdata/compassql/issues/260)
```
kanitw: Don't just put issue url.  Describe briefly what you should FIX too.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+  let specQ: SpecQuery = {mark: splitShorthand[0], encodings: [] as EncodingQuery[]};
+
+  for (let i = 1; i < splitShorthand.length; i++) {
+    let part = splitShorthand[i];
+    const splitPart = splitWithTail(part, ':', 1);
+    const splitPartKey = splitPart[0];
+    const splitPartValue = splitPart[1];
+
+    if (CHANNEL_INDEX[splitPartKey] || splitPartKey === '?') {
+      shorthandParser.encoding(specQ, splitPartKey, splitPartValue);
```
kanitw: Since you're not using output from shorthandParser.encoding to determine whether you can skip the rest of stuff, I wonder why you have to pass `specQ` inside?

You can just return `encQ` and do 

```
const encQ =  encoding(specQ, splitPartKey, splitPartValue)
specQ.encodings.push(encQ);
```

This way, your test code will be way better abstracted.

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+      continue;
+    }
+
+    if (splitPartKey === 'calculate') {
+      shorthandParser.calculate(specQ, splitPartValue);
```
kanitw: No need to pass specQ inside, since you're not using output from shorthandParser.encoding to determine whether you can skip the rest of stuff

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+      continue;
+    }
+
+    if (splitPartKey === 'filter') {
+      shorthandParser.filter(specQ, splitPartValue);
```
kanitw: No need to pass specQ inside, since you're not using output from shorthandParser.encoding to determine whether you can skip the rest of stuff

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+      continue;
+    }
+
+    if (splitPartKey === 'filterInvalid') {
+      shorthandParser.filterInvalid(specQ, splitPartValue);
```
kanitw: No need to pass specQ inside, since you're not using output from shorthandParser.encoding to determine whether you can skip the rest of stuff

kanitw: This method actually doesn't do anything special inside, no need to extract them.  Just call them here.

src/query/shorthand.ts
kanitw: This method actually doesn't do anything special inside, no need to extract them.  Just call them here.

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,159 @@ describe('query/shorthand', () => {
...
+
+    describe('filter', () => {
+      it('should correctly parse filter from shorthand', () => {
+        let specQ: SpecQuery = {} as SpecQuery;
+        let shorthandPart = 'filter:[{"field":"color","equal":"red"},"datum[\\"b2\\"] > 60",{"field":"color","oneOf":["red","yellow"]},{"field":"x","range":[0,5]}]';
+        let splitPartValue = splitWithTail(shorthandPart, ':', 1)[1];
+
+        shorthandParser.filter(specQ, splitPartValue);
+        assert.deepEqual(specQ.transform.filter, [
+          {field: 'color', equal: 'red'}, 'datum["b2"] > 60', {field: 'color', oneOf: ['red', 'yellow']}, {field: 'x', range: [0,5]}
```
kanitw: You can't really specify filter as `datum["b2"] > 60` in JSON?

It has to be one of
- `datum.b2 > 60`
- `datum['b2'] > 60`
- `datum[\"b2\"] > 60`

test/query/shorthand.test.ts
```diff
@@ -432,16 +572,16 @@ describe('query/shorthand', () => {
        assert.equal(str, 'bin(a,q,maxbins=20)');
     });
 
-    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType log', () => {
+    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType ordinal', () => {
       const str = fieldDefShorthand({
-        channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, bin: {maxbins: 20}, scale: {type: ScaleType.LOG}
+        channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, bin: {maxbins: 20}, scale: {type: ScaleType.ORDINAL}
```
kanitw: Bin with ordinal scale?  We don't support it.   I don't understand why you just don't put an obviously valid one like LINEAR here.

test/query/shorthand.test.ts
```diff
@@ -297,34 +452,19 @@ describe('query/shorthand', () => {
...
 
-    it('should return correct fieldDefShorthand string for bin with maxbins, axis with orient, scale with scaleType ' +
-       'and sort field definition object', () => {
+    it('should return correct fieldDefShorthand string for bin with maxbins, axis with orient, scale with scaleType ', () => {
       const str = fieldDefShorthand({
-        axis: {orient: AxisOrient.TOP}, bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LOG}, sort: {field: 'a', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}
+        axis: {orient: AxisOrient.TOP}, bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LOG}
       });
-      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"log"},sort={"field":"a","op":"mean","order":"descending"},axis={"orient":"top"})');
+      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"log"},axis={"orient":"top"})');
```
kanitw: Bin wit log scale

test/query/shorthand.test.ts
```diff
@@ -297,34 +452,19 @@ describe('query/shorthand', () => {
...
-    });
-
     it('should return correct fieldDefShorthand string for bin with maxbins, scale with scaleType ' +
        'and sort field definition object', () => {
       const str = fieldDefShorthand({
-        bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LOG}, sort: {field: 'a', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}
+        bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LOG}
       });
-      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"log"},sort={"field":"a","op":"mean","order":"descending"})');
+      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"log"})');
```
kanitw: Bin with log scale

test/query/shorthand.test.ts
```diff
@@ -432,16 +572,16 @@ describe('query/shorthand', () => {
...
       });
-      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"log"})');
+      assert.equal(str, 'bin(a,q,maxbins=20,scale={"type":"ordinal"})');
     });
 
-    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType log when only field, bin, and type are included', () => {
+    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType ordinal when only field, bin, and type are included', () => {
       const str = fieldDefShorthand({
-        channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, bin: {maxbins: 20}, scale: {type: ScaleType.LOG}
+        channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, bin: {maxbins: 20}, scale: {type: ScaleType.ORDINAL}
```
kanitw: Bin with ordinal scale

src/query/shorthand.ts
```diff
@@ -277,3 +282,205 @@ export function fieldDef(encQ: EncodingQuery,
...
+      const encQ = shorthandParser.encoding(splitPartKey, splitPartValue);
+      specQ.encodings.push(encQ);
+      continue;
+    }
+
+    if (splitPartKey === 'calculate') {
+      let transformQ: TransformQuery = specQ.transform || {};
+
+      let calculate = [];
+      let formulas = splitPartValue.split(',');
```
kanitw: Actually, fixing the shorthand method for calculate to be correct (#260) would take < 10 mins.  
How about you send another PR to fix that and update this one (parse) to be correct -- instead of leaving it like this.

test/query/shorthand.test.ts
```diff
@@ -297,34 +427,19 @@ describe('query/shorthand', () => {
       assert.equal(str, 'a,q,sort={"field":"a","op":"mean","order":"descending"}');
     });
 
-    it('should return correct fieldDefShorthand string for bin with sort field definition object', () => {
```
kanitw: You're changing code for parsing -- I don't understand why you're removing code for shorthand generation.  
If the reason is the old example was wrong, the solution is definite NOT to remove them.

test/query/shorthand.test.ts
```diff
@@ -297,34 +427,19 @@ describe('query/shorthand', () => {
...
-        bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, sort: {field: 'a', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}
-      });
-      assert.equal(str, 'bin(a,q,maxbins=20,sort={"field":"a","op":"mean","order":"descending"})');
-    });
-
     it('should return correct fieldDefShorthand string for bin with maxbins, scale with scaleType ' +
        'and sort field definition object', () => {
       const str = fieldDefShorthand({
-        bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LOG}, sort: {field: 'a', op: AggregateOp.MEAN, order: SortOrder.DESCENDING}
+        bin: {maxbins: 20}, channel: Channel.X, field: 'a', type: Type.QUANTITATIVE, scale: {type: ScaleType.LINEAR}
```
kanitw: Now the test description says sort but the test doesn't contain sort.   Update the description accordingly.  
Also make sure we still have a case where we test ordinal scale with sort definition object.  I believe that's the original intention of this test case.

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,134 @@ describe('query/shorthand', () => {
...
+    });
+
+    it('should correctly split a string when `count` is greater than the number of delimiters in the string', () => {
+      let result = splitWithTail('012-345', '-', 3);
+      assert.deepEqual(result, ['012', '345', '', '']);
+    });
+  });
+
+  describe('parse', () => {
+    it('should correctly parse a shorthand string', () => {
```
kanitw: append "with calculate, filter, and filterInvalid"

test/query/shorthand.test.ts
```diff
@@ -33,6 +35,134 @@ describe('query/shorthand', () => {
...
+        },
+        mark: Mark.POINT,
+        encodings: [
+          {channel: Channel.X, field: 'b2', type: Type.QUANTITATIVE},
+          {bin: {}, channel: Channel.Y, field: 'balance', type: Type.QUANTITATIVE}
+        ]
+      });
+    });
+
+    it('should correctly parse an ambiguous shorthand with aggregate and bin as enum spec', () => {
```
kanitw: "and with hasFn"

test/query/shorthand.test.ts
```diff
@@ -432,16 +547,16 @@ describe('query/shorthand', () => {
        assert.equal(str, 'bin(a,q,maxbins=20)');
     });
 
-    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType log', () => {
+    it('should return correct fieldDefShorthand string for bin field with maxbins and scale with scaleType ordinal', () => {
```
kanitw: Below is obviously not ordinal

test/query/shorthand.test.ts
felixcodes: The old example was incorrect and its only purpose was to demonstrate bin with a nested parent property. Since we already have bin and scale test I thought it was unnecessary and removed it.

src/query/shorthand.ts
```diff
@@ -300,3 +303,200 @@ export function fieldDef(encQ: EncodingQuery,
...
+      continue;
+    }
+
+    if (splitPartKey === 'calculate') {
+      let transformQ: TransformQuery = specQ.transform || {};
+      let calculate: Formula[] = [];
+      let fieldAndExpr = JSON.parse(splitPartValue);
+
+      for (let field in fieldAndExpr) {
+        let formula: Formula = {} as Formula;
```
kanitw: why not just 
```
calculate.push({field: field, expr: fieldAndExpr[field]})  
```

src/query/shorthand.ts
```diff
@@ -300,3 +303,200 @@ export function fieldDef(encQ: EncodingQuery,
...
+    if (CHANNEL_INDEX[splitPartKey] || splitPartKey === '?') {
+      const encQ = shorthandParser.encoding(splitPartKey, splitPartValue);
+      specQ.encodings.push(encQ);
+      continue;
+    }
+
+    if (splitPartKey === 'calculate') {
+      let transformQ: TransformQuery = specQ.transform || {};
+      let calculate: Formula[] = [];
+      let fieldAndExpr = JSON.parse(splitPartValue);
```
kanitw: rename to `fieldExprMapping`

src/query/shorthand.ts
```diff
@@ -300,3 +303,200 @@ export function fieldDef(encQ: EncodingQuery,
...
+    const splitPartValue = splitPart[1];
+
+    if (CHANNEL_INDEX[splitPartKey] || splitPartKey === '?') {
+      const encQ = shorthandParser.encoding(splitPartKey, splitPartValue);
+      specQ.encodings.push(encQ);
+      continue;
+    }
+
+    if (splitPartKey === 'calculate') {
+      let transformQ: TransformQuery = specQ.transform || {};
```
kanitw: Remove all declaration of `transformQ`

Just say `specQ.transform = specQ.transform || {}` and always use `specQ.transform`

This way you don't have to assign it back to specQ.transform later, which is redundant.

PR #321, opened 2017-02-06
Add scalePropertiesSupportedByChannel Constraint

### LINKED ISSUE #174: scalePropertiesSupportedByChannel Constraint
Not all scale properties are supported by all encoding channels.  
We should add a constraint to prevent outputting unsupported values. 

- [ ] We can leverage `channelScalePropertyIncompatability` in Vega-Lite

mattwchun:
Fix #174. Adds another encoding constraint because not all scale properties are supported by all encoding channels yet.

### DIFF
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -246,13 +246,12 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
     strict: true,
     satisfy: (encQ: EncodingQuery) => {
       if (encQ) {
-        let channel = encQ.channel as Channel;
-        let scale = encQ.scale;
+        let channel: Channel = encQ.channel as Channel;
+        let scale: ScaleQuery = encQ.scale as ScaleQuery;
         if (channel && scale) {
-          let scaleProps = Object.keys(scale);
-          for (let scaleProp in scaleProps) {
+          for (let scaleProp in scale) {
 
-            if (!scaleProps.hasOwnProperty(scaleProp)) continue;
+            if (!scale.hasOwnProperty(scaleProp)) continue;
 
             if (scaleProp === 'type' || scaleProp === 'name' || scaleProp === 'enum') {
               // ignore type and properties of wildcards
--- a/src/constraint/encoding.ts
+++ b/src/constraint/encoding.ts
@@ -1,6 +1,6 @@
 import {AggregateOp} from 'vega-lite/src/aggregate';
 import {Channel, getSupportedRole} from 'vega-lite/src/channel';
-import {ScaleType, scaleTypeSupportProperty, hasDiscreteDomain} from 'vega-lite/src/scale';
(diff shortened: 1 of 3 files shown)


src/constraint/encoding.ts
```diff
@@ -239,6 +239,28 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+    properties: [Property.CHANNEL, Property.SCALE],
+    allowWildcardForProperties: false, // unsure about this
+    strict: true,
+    satisfy: (encQ: EncodingQuery) => {
+      if (encQ) {
+        let channel = encQ.channel as Channel;
+        let scale = encQ.scale;
+        if (channel && scale) {
+          let scaleProps = Object.keys(scale);
+          for (let scaleProp in scaleProps) {
```
kanitw: `scaleProp in scale` is fine too.  

But you probably wanna use `scale.hasOwnProperty()` check like typical object key loop in JS.
More importantly, check if the prop is actually scale property (not wildcard's `name` and `enum`)

src/constraint/encoding.ts
```diff
@@ -239,6 +239,35 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
       return true;
     }
   },{
+    name: 'scalePropertiesSupportedByChannel',
+    description: 'Not all scale properties are supported by all encoding channels',
+    properties: [Property.CHANNEL, Property.SCALE],
```
kanitw: properties here should include scale properties too, otherwise they won't get triggered.

src/constraint/encoding.ts
```diff
@@ -239,6 +239,35 @@ export const ENCODING_CONSTRAINTS: EncodingConstraintModel[] = [
...
+    allowWildcardForProperties: false, // unsure about this
```
kanitw: I'd make it true :)

kanitw: (But then manually check if channel is not wildcard.  

(If channel is a wildcard, there is no point checking about this at all)

kanitw:
@mattwchun  Looks like you haven't add a test for it. :)

test/constraint/encoding.test.ts
```diff
@@ -634,6 +634,189 @@ describe('constraints/encoding', () => {
...
+        type: '?',
+        scale: {
+          paddingInner: 20
+        }
+      };
+
+      assert.isTrue(ENCODING_CONSTRAINT_INDEX['scalePropertiesSupportedByChannel'].satisfy(encQ, schema, new PropIndex<Wildcard<any>>(), defaultOpt));
+    });
+
+    it('incompatible scale property range with channel x', () => {
```
kanitw: `it('should return false when using range with channel x', ...`

kanitw:
@mattwchun  Thanks :)

PR #325, opened 2017-02-07
[WIP] Adding index to fieldSchema for easy comparison	

kanitw:


kanitw:
The master already have these index code

PR #354, opened 2017-03-07
Split constraint files

AkshatSh:
Updated constraint encodings to be split across

- field.ts with the field constraints
- value.ts with the value constraints

The generic constraint checker, constraint model, etc were moved to base.ts

Only the check encoding function is left in encoding.ts


test/constraint/encoding.test.ts
```diff
@@ -9,7 +9,9 @@ import {Type} from 'vega-lite/build/src/type';
 import {Property} from '../../src/property';
 import {PropIndex} from '../../src/propindex';
 import {DEFAULT_QUERY_CONFIG} from '../../src/config';
-import {EncodingConstraintModel, FIELD_CONSTRAINTS, FIELD_CONSTRAINT_INDEX, VALUE_CONSTRAINT_INDEX} from '../../src/constraint/encoding';
```
kanitw: should we also split this file to provide 1:1 mappings between src and test?

PR #359, opened 2017-03-09
Fix logic of the "omitTableWithOcclusionIfAutoAddCount

yhoonkim:
When I try the below query, it returns `Uncaught TypeError: Cannot read property 'bin' of undefined`.  (I used `cars.json`.)
```json
{
  "spec": {
    "mark": "point",
    "encodings": [
      {
        "channel": "?",
        "field": "Acceleration", 
        "type": "quantitative",
        "bin": "?"
      }
    ]
  },
  "config": {
    "autoAddCount": true,
    "overlay": {
      "line": false
    }
  }
};
```
I guess it's because of the constraint doesn't check whether `yEncQ` and `xEncQ` are undefined. So I changed it. 

In addition, I think this constraint is still valid for 1D case. If it is, will it be better to add a new constraint?

### DIFF
--- a/src/constraint/spec.ts
+++ b/src/constraint/spec.ts
@@ -606,8 +606,8 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
           if (specM.isAggregate()) { // TODO: refactor based on profiling statistics
             const xEncQ = specM.getEncodingQueryByChannel(Channel.X);
             const yEncQ = specM.getEncodingQueryByChannel(Channel.Y);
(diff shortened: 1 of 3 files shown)


kanitw:
Thanks.  A better commit message and PR title would be "Fix logic of the 
"omitTableWithOcclusionIfAutoAddCount" constraint.

I'll take a closer look to see if this is the optimal fix.

kanitw:
@yhoonkim I replaced your commit with a few relevant commits.  

Please rebase and merge if you think this makes sense + the test passes

PR #365, opened 2017-03-29
Enumerate sort.field with '*'

yhoonkim:
I added `autoAddCountOnSort` on `config` to enable to enumerate `sort.field` with '*'.


src/wildcard.ts
```diff
@@ -279,6 +279,9 @@ export const DEFAULT_ENUM_INDEX: EnumIndex = {
 export function getDefaultEnumValues(prop: Property, schema: Schema, opt: QueryConfig): any[] {
   if (prop === 'field' || (isEncodingNestedProp(prop) && prop.parent === 'sort' && prop.child === 'field')) {
     // For field, by default enumerate all fields
+    if ((isEncodingNestedProp(prop) && prop.parent === 'sort' && prop.child === 'field') && opt.autoAddCountOnSort){
```
kanitw: The insertion here is a bit weird because it makes the following comment outdated:
```
For field, by default enumerate all fields
```

The name `autoAddCountOnSort` is a bit misleading since you are not really adding count here, you are adding `'*'`.   

I'd say remove the misleading `autoAddCountOnSort` config and just keep this logic here, but add proper comment describing why you need to add '*' to `sort.field` (to support count).

src/constraint/encoding.ts
```diff
@@ -375,6 +375,18 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
       }
       return true;
     }
+  },{
+    name: 'onlyCountWithAutoCountFieldOnSort',
```
kanitw: This isn't related to autoCount.  

`onlyUseCountWithAsteriskSortField`?

src/constraint/encoding.ts
```diff
@@ -375,6 +375,18 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
...
+    description: 'Sort by * if and only if op is count',
+    properties: [Property.SORT, getEncodingNestedProp('sort', 'field'), getEncodingNestedProp('sort', 'op')],
+    allowWildcardForProperties: false,
+    strict: true,
+    satisfy: (fieldQ: FieldQuery, _: Schema, __: PropIndex<Wildcard<any>>, ___: QueryConfig) => {
+      if (fieldQ.sort && !!(fieldQ.sort as SortField).field) {
```
kanitw: Don't hack using `as SortField` -- do proper type guard using `isSortField` in VL's sort.ts

test/constraint/encoding.test.ts
```diff
@@ -492,6 +492,25 @@ describe('constraints/encoding', () => {
    });
   });
 
+  describe('onlyCountWithAutoCountFieldOnSort', () => {
```
kanitw: rename

kanitw:
- [ ]  Improve test coverage

kanitw:
There are some conflicts, possibly because encoding.ts and its test file have been splitted.

yhoonkim:
@kanitw 
Current master branch fails on testing.
`TypeError: Cannot read property 'forEach' of undefined`
 I'll merge it after master branch resolves this error.

kanitw:
@yhoonkim Can you rebase this when you have time?

yhoonkim:
@kanitw I rebased this branch but still tests fail.

PR #369, opened 2017-04-12
Fix#: 364 added AutoCountQUery

AkshatSh:
Added auto count query throughout code base. 

I labeled any questionable changes with TODO(akshatsh) in a comment above and gave a description about what the potential problem is. 

Fixed: https://github.com/vega/compassql/issues/364

### DIFF
--- a/src/constraint/field.ts
+++ b/src/constraint/field.ts
@@ -11,7 +11,7 @@ import {isWildcard, Wildcard} from '../wildcard';
 import {PrimitiveType, Schema} from '../schema';
 import {contains} from '../util';
 
-import {scaleType, FieldQuery, ScaleQuery, toFieldDef} from '../query/encoding';
+import {scaleType, FieldQuery, ScaleQuery, toFieldDef, AutoCountQuery, isAutoCountQuery} from '../query/encoding';
 import {EncodingConstraintModel, EncodingConstraint} from './base';
 
 export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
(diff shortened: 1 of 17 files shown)


src/constraint/field.ts
```diff
@@ -96,9 +96,10 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
     properties: [Property.AGGREGATE, Property.AUTOCOUNT, Property.TIMEUNIT, Property.BIN],
     allowWildcardForProperties: true,
     strict: true,
-    satisfy: (fieldQ: FieldQuery, _: Schema, __: PropIndex<Wildcard<any>>, ___: QueryConfig) => {
+    // TODO(akshatsh): Allow field query with optional autocount?
```
kanitw: Yes, this seems reasonable.

src/constraint/field.ts
```diff
@@ -96,9 +96,10 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
...
+    satisfy: (fieldQ: FieldQuery & AutoCountQuery, _: Schema, __: PropIndex<Wildcard<any>>, ___: QueryConfig) => {
```
kanitw: `FieldQuery | AutoCountQuery`, not `FieldQuery & AutoCountQuery`.

Same for a few other places.

src/constraint/spec.ts
```diff
@@ -138,16 +138,21 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
         return every(specM.getEncodings(), (encQ: EncodingQuery) => {
-          // TODO(akshatsh): should value query return false?
           if (isValueQuery(encQ)) {return false;}
-          if (encQ.autoCount !== undefined) {
+
+          if (isAutoCountQuery(encQ)) {
             return true;
           }
+
+          /*if (encQ.autoCount !== undefined) {
```
kanitw: This seems reasonable, so you can remove the commented code.

src/constraint/spec.ts
```diff
@@ -174,11 +179,13 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
           // (3) nominal or ordinal field
           // or at least have potential to be (still ambiguous).
           return some(specM.getEncodings(), (encQ: EncodingQuery) => {
-            if (isFieldQuery(encQ) && encQ.type === Type.QUANTITATIVE) {
-              if (encQ.autoCount === false) {
+            if ((isFieldQuery(encQ) || isAutoCountQuery(encQ)) && encQ.type === Type.QUANTITATIVE) {
+
+              // TODO(akshatsh): autoCount undefined okay?
```
kanitw: Yep. If `autoCount` is `undefined`, then it is not an AutoCountQuery

src/constraint/spec.ts
```diff
@@ -265,9 +274,10 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
       if (specM.isAggregate()) {
         let hasNonFacetDim = false, hasDim = false, hasEnumeratedFacetDim = false;
         specM.specQuery.encodings.forEach((encQ, index) => {
-          if (isValueQuery(encQ) || encQ.autoCount === false) return; // skip unused field
+          if (isValueQuery(encQ) || (isAutoCountQuery(encQ) && encQ.autoCount === false)) return; // skip unused field
```
kanitw: We have so many `(isAutoCountQuery(encQ) && encQ.autoCount === false)`. 

I wonder if we should add `isDisabledAutoCountQuery` to make it concise.

Similarly, maybe also `isEnabledAutoCountQuery` for  `(isAutoCountQuery(encQ) && encQ.autoCount === true)`.

kanitw:
Looks like  there are some conflicts -- rebase?

kanitw:
@AkshatSh  I push misc. fixes -- can you review and see if you agree with my changes. Feel free to comment if there are parts you disagree. 
(It would be useful to look at *both* my single commit and how it differs from master.)

Also, you'll probably want to rebase.

PR #370, opened 2017-04-14
temporal field should be able to be a group-by axis

yhoonkim:
Because `isContinuous()` function in Vega-Lite returns `true` if temporal field has a non-periodic `timeUnit`, `hasAppropriateGraphicTypeForMark` rejects a spec having that temporal field and an aggregated quantitative field.

I changed `isMeasure` to be `false` when a field is temporal. (I thought a temporal field with any `timeUnit` can draw a line/area chart)

I couldn't add a test for this PR because master branch is broken so that I couldn't run `npm run test`.
Here is the case that current master is not working well:
```json
{
    "spec": {
      "mark": "line",
      "encodings": [
        {
          "channel": "?",
          "field": "Origin",
          "type": "nominal"
        },
        {
          "channel": "?",
          "field": "Accerleration",
          "type": "quantitative",
          "aggregate": "?"
        },
        {
          "channel": "?",
          "field": "Year",
          "type": "temporal",
          "timeUnit": "yearmonthdate"
        }
      ]
    },
    "config": {
      "autoAddCount": false
    }
  };
```

### DIFF
--- a/test/constraint/spec.test.ts
+++ b/test/constraint/spec.test.ts
@@ -329,7 +329,7 @@ describe('constraints/spec', () => {
     });
   });
 
-  describe('hasAppropriateGraphicTypeForMark', () => {
+  describe.only('hasAppropriateGraphicTypeForMark', () => {
     describe('bar, tick', () => {
       it('should return true for plots with one dimension and one measure on x/y', () => {
         [Mark.BAR, Mark.TICK].forEach((mark) => {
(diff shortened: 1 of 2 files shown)


test/constraint/spec.test.ts
```diff
@@ -329,7 +329,7 @@ describe('constraints/spec', () => {
     });
   });
 
-  describe('hasAppropriateGraphicTypeForMark', () => {
+  describe.only('hasAppropriateGraphicTypeForMark', () => {
```
yhoonkim: This test has not been run yet.

test/constraint/spec.test.ts
```diff
@@ -464,6 +464,19 @@ describe('constraints/spec', () => {
           assert.isFalse(SPEC_CONSTRAINT_INDEX['hasAppropriateGraphicTypeForMark'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
         });
       });
+
+      it('should return true for aggregate line/area with one temporal field and one continuous field', () => {
```
yhoonkim: This test has not been run actually due to the broken test environment.

kanitw:
@yhoonkim I think this works in master now.  Can you check again?

kanitw:
I'm pretty sure it works now.

PR #371, opened 2017-04-23
[WIP] Boxplot

mattwchun:
- [ ] Update CompassQL Mark Type and Aggregation Function

- [ ] Update default aggregation to enumerate [undefined, 'boxplot'] instead of [undefined, 'mean']

- [ ] Add New Constraints
  - [ ] one axis should be continuous and one should be discrete
  - [ ] aggregate = boxplot, use mark = boxplot

- [ ] Update Rankings

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -4,7 +4,7 @@ import {Encoding} from 'vega-lite/build/src/encoding';
 import {FieldDef} from 'vega-lite/build/src/fielddef';
 import {Mark} from 'vega-lite/build/src/mark';
 import {Type} from 'vega-lite/build/src/type';
-import {FacetedUnitSpec} from 'vega-lite/build/src/spec';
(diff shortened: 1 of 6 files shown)


mattwchun:
test cases

- [ ] write a query with mark = ?  with one Q-field where aggregate = {enum: [undefined, boxplot])
- [ ] write a query with mark = ?  with one N-field and one Q-field  where aggregate = {enum: [undefined, boxplot])
- [ ] write a query with mark = ?  with two Q-fields  where aggregate = {enum: [undefined, boxplot]) -- this one shouldn't show any boxplot because it's dumb

- [ ] write a query with mark = ?  with two Q-fields  where aggregate = {enum: [undefined, boxplot]), bin = {enum: [false, true]) -- the case where one of them is binned would make sense

PR #372, opened 2017-05-16
Fix #70 high cardinality nomial fields distinguished as new type

### LINKED ISSUE #70: Distinguish high-cardinality strings from nominal fields
Fields with too high cardinality takes up a lot of space and can be slow to render.

- [ ] add a flag `isKeyLike` (or some better name) to schema

We might want to consider a few options:
- [ ] distinguish between categories (low cardinality) and text (high cardinality) as they serve different purpose in data analysis anyway. 
   - [ ] Check if the cardinality is above X% (50%?) of the overall data count and above minimum threshold (e.g., 40)

Maybe check if "if the cardinality is above ~80% of the overall data count"  or some similar criteria

- [ ] Add a constraint that excludes fields with too high cardanality from being added automatically.

AkshatSh:
Added a new type for the high cardinality strings (ExpandedType.KEY)
Added constants for determining type to configuration
- I think the enumeration was handled properly, not entirely sure though

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -70,6 +70,9 @@ export interface QueryConfig {
   // EFFECTIVENESS PREFERENCE
   maxGoodCardinalityForColor?: number; // FIXME: revise
   maxGoodCardinalityForFacet?: number; // FIXME: revise
+  // HIGH CARDINALITY STRINGS
+  percentUnique?: number;
+  minimumKeys?: number;
 }
 
 export const DEFAULT_QUERY_CONFIG: QueryConfig = {
@@ -126,6 +129,10 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   // RANKING PREFERENCE
   maxGoodCardinalityForFacet: 5, // FIXME: revise
   maxGoodCardinalityForColor: 7, // FIXME: revise
+
+  // HIGH CARDINALITY STRINGS
+  percentUnique: .8,
+  minimumKeys: 50,
 };
 
 export function extendConfig(opt: QueryConfig) {
--- a/src/config.ts
+++ b/src/config.ts
@@ -71,8 +71,8 @@ export interface QueryConfig {
   maxGoodCardinalityForColor?: number; // FIXME: revise
   maxGoodCardinalityForFacet?: number; // FIXME: revise
   // HIGH CARDINALITY STRINGS
-  percentUnique?: number;
-  minimumKeys?: number;
+  minPercentUniqueForKey?: number;
+  minCardinalityForKey?: number;
 }
 
 export const DEFAULT_QUERY_CONFIG: QueryConfig = {
@@ -131,8 +131,8 @@ export const DEFAULT_QUERY_CONFIG: QueryConfig = {
   maxGoodCardinalityForColor: 7, // FIXME: revise
 
   // HIGH CARDINALITY STRINGS
-  percentUnique: .8,
-  minimumKeys: 50,
+  minPercentUniqueForKey: .8,
+  minCardinalityForKey: 50,
 };
 
 export function extendConfig(opt: QueryConfig) {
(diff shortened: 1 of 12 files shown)


src/config.ts
```diff
@@ -70,6 +70,9 @@ export interface QueryConfig {
   // EFFECTIVENESS PREFERENCE
   maxGoodCardinalityForColor?: number; // FIXME: revise
   maxGoodCardinalityForFacet?: number; // FIXME: revise
+  // HIGH CARDINALITY STRINGS
+  percentUnique?: number;
```
kanitw: minPercentUniqueForKey, minCardinalityForKey

src/constraint/field.ts
```diff
@@ -22,7 +23,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
     strict: true,
     satisfy: (fieldQ: FieldQuery, _: Schema, __: PropIndex<Wildcard<any>>, ___: QueryConfig) => {
       if (fieldQ.aggregate) {
-        return fieldQ.type !== Type.ORDINAL && fieldQ.type !== Type.NOMINAL;
+        return fieldQ.type !== Type.ORDINAL && fieldQ.type !== Type.NOMINAL && fieldQ.type !== ExpandedType.KEY;
```
kanitw: You might wanna add `isDiscrete` to `expandedtype.ts` and use `!isDiscrete(fieldQ.type)`

src/constraint/field.ts
```diff
@@ -300,7 +301,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
         const type = fieldQ.type;
         const sType = scaleType(fieldQ);
 
-        if (contains([Type.ORDINAL, Type.NOMINAL], type)) {
+        if (contains([Type.ORDINAL, Type.NOMINAL, ExpandedType.KEY], type)) {
```
kanitw: isDiscrete :)

src/constraint/spec.ts
```diff
@@ -613,8 +615,10 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
             return xEncQ && yEncQ && (xIsMeasure !== yIsMeasure) &&
               // and the dimension axis should not be nominal
               // TODO: make this clause optional
-              !(isFieldQuery(xEncQ) && !xIsMeasure && xEncQ.type === Type.NOMINAL) &&
-              !(isFieldQuery(yEncQ) && !yIsMeasure && yEncQ.type === Type.NOMINAL)
+
+              // TODO: should this one consider ORDINAL too.
+              !(isFieldQuery(xEncQ) && !xIsMeasure && (xEncQ.type === Type.NOMINAL || xEncQ.type === ExpandedType.KEY)) &&
```
kanitw: use isDiscrete and remove the TODO

kanitw:
Also, make sure to convert type for output methods

- [ ] `toFieldDef` in encoding.ts
- [ ] model.ts's `toShorthand` and `toSpec`

src/schema.ts
```diff
@@ -120,6 +121,12 @@ export function build(data: any,  tableSchema: TableSchema<TableSchemaFieldDescr
       vlType = VLType.NOMINAL;
     }
 
+    if (vlType === VLType.NOMINAL 
+      && distinct / fieldProfile.count > DEFAULT_QUERY_CONFIG.minPercentUniqueForKey
```
kanitw: Reading directly from `DEFAULT_QUERY_CONFIG` here means that whatever config we provide wouldn't be used.  

We already call `opt = extend({}, DEFAULT_QUERY_CONFIG, opt);` at the beginning of this method, so better read the config from `opt`.

test/schema.test.ts
```diff
@@ -3,6 +3,8 @@ import {assert} from 'chai';
 import {Type} from 'vega-lite/build/src/type';
 import {Channel} from 'vega-lite/build/src/channel';
 
+import {ExpandedType} from '../src/query/ExpandedType';
```
kanitw: Why import here, seems like it's not used?

test/constraint/spec.test.ts
```diff
@@ -380,7 +380,7 @@ describe('constraints/spec', () => {
             ]
           });
           assert.isFalse(SPEC_CONSTRAINT_INDEX['hasAppropriateGraphicTypeForMark'].satisfy(specM, schema, DEFAULT_QUERY_CONFIG));
-        });
+      });
```
kanitw: why changing tab here?

PR #376, opened 2017-05-23
Provide SpecQueryGroup Generic.

### LINKED ISSUE #375: Make SpecQueryModelGroup a Generic
Goal: Create a `SpecQueryModelGroup` - like generic so that we may parameterize the type stored in `_items`. This will allow us to create a `SpecQueryPlotOptionGroup` to be used in Voyager as a serializable return value for `recommend`.

See https://github.com/vega/voyager/issues/330 for more info.

vlandham:
SpecQueryModelGroup Is based on this class.

closes #375

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -346,14 +346,15 @@ export class SpecQueryModel {
   }
 }
 
-export class SpecQueryModelGroup {
+export class SpecQueryGroup<T> {
+
   private _name: string;
   private _path: string;
-  private _items: (SpecQueryModel | SpecQueryModelGroup)[];
+  private _items: (SpecQueryGroup<T> | T)[];
   private _groupBy: GroupBy;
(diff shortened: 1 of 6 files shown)


src/model.ts
```diff
@@ -362,10 +363,14 @@ export class SpecQueryModelGroup {
     this._orderGroupBy = orderGroupBy;
   }
 
-  public getTopSpecQueryModel(): SpecQueryModel {
+  public isItemSpecQueryGroup(item: SpecQueryGroup<T> | T): item is SpecQueryGroup<T> {
```
kanitw: This looks more like static method? (Maybe just extract it outside of the class so it can be reused easily?)

vlandham:
@kanitw - sounds good. the code is updated to have `isItemSpecQueryGroup` as a separate exported function.

Issue #377, opened 2017-05-24
Remove vega-lite types from compassql 


vlandham:
The types are used the import of `vega-lite` utils - but should not be duplicated in this repo.

PR #378, opened 2017-05-24
Remove vega-util types. Make type imports work from node_modules

vlandham:
@kanitw I believe the issue was that `node_modules` was excluded from the type imports. 
With this removed (the rest of the excludes were not relevant to the project), the build works for me.


vlandham:
closes #377

PR #379, opened 2017-06-13
Make Schema conform to Table Schema Spec.

vlandham:
This changes the `Schema` interface to be compliant with the Table Schema spec. 

https://specs.frictionlessdata.io/table-schema/

As discussed in https://github.com/vega/voyager/issues/338, here we:

* provide interfaces for `TableSchemaFieldDescriptor` and `TableSchema`
* rename field to name - the required ID of the field
* rename type to vlType - as this is additional data not captured in table schema
* rename primitiveType to type
* rename primitiveType date to datetime.

This introduces some inconsistencies between this new `Schema` and (for example) the `FieldQuery` interface (field === name). we can work toward changing more interfaces if this is a problem.

### DIFF
--- a/src/constraint/field.ts
+++ b/src/constraint/field.ts
@@ -7,7 +7,7 @@ import {QueryConfig} from '../config';
 import {getEncodingNestedProp, Property, SCALE_PROPS} from '../property';
 import {PropIndex} from '../propindex';
 import {isWildcard, Wildcard} from '../wildcard';
-import {PrimitiveType, Schema} from '../schema';
+import {DataType, Schema} from '../schema';
 import {contains} from '../util';
 
 import {scaleType, FieldQuery, ScaleQuery, toFieldDef} from '../query/encoding';
@@ -208,7 +208,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
         return true;
       }
 
-      const primitiveType = schema.primitiveType(fieldQ.field as string);
+      const primitiveType = schema.type(fieldQ.field as string);
       const type = fieldQ.type;
 
       if (!encWildcardIndex.has('field') && !encWildcardIndex.has('type') && !opt.constraintManuallySpecifiedValue) {
@@ -217,13 +217,13 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
       }
 
       switch (primitiveType) {
-        case PrimitiveType.BOOLEAN:
-        case PrimitiveType.STRING:
+        case DataType.BOOLEAN:
+        case DataType.STRING:
           return type !== Type.QUANTITATIVE && type !== Type.TEMPORAL;
-        case PrimitiveType.NUMBER:
-        case PrimitiveType.INTEGER:
+        case DataType.NUMBER:
+        case DataType.INTEGER:
           return type !== Type.TEMPORAL;
-        case PrimitiveType.DATE:
+        case DataType.DATETIME:
           // TODO: add NOMINAL, ORDINAL support after we support this in Vega-Lite
           return type === Type.TEMPORAL;
         case null:
@@ -249,7 +249,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
         return fieldQ.type === Type.QUANTITATIVE;
       }
 
-      return schema.type(fieldQ.field as string) === fieldQ.type;
+      return schema.vlType(fieldQ.field as string) === fieldQ.type;
     }
   },{
    name: 'maxCardinalityForCategoricalColor',
--- a/src/constraint/field.ts
+++ b/src/constraint/field.ts
@@ -7,7 +7,7 @@ import {QueryConfig} from '../config';
 import {getEncodingNestedProp, Property, SCALE_PROPS} from '../property';
 import {PropIndex} from '../propindex';
 import {isWildcard, Wildcard} from '../wildcard';
-import {DataType, Schema} from '../schema';
+import {PrimitiveType, Schema} from '../schema';
 import {contains} from '../util';
 
 import {scaleType, FieldQuery, ScaleQuery, toFieldDef} from '../query/encoding';
@@ -208,7 +208,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
         return true;
       }
 
-      const primitiveType = schema.type(fieldQ.field as string);
+      const primitiveType = schema.primitiveType(fieldQ.field as string);
       const type = fieldQ.type;
 
       if (!encWildcardIndex.has('field') && !encWildcardIndex.has('type') && !opt.constraintManuallySpecifiedValue) {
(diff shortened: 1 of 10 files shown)


kanitw:
Wow this is fast -- looks like the test is failing though.   

I'll take a look after I arrive at UW (still commuting)

vlandham:
haha. yeah, i was working on it - hoping it would be agreed upon ;) - and yes - looks like i was mistaken about where compassql was running tests. will fix that now.

src/schema.ts
```diff
@@ -131,44 +164,44 @@ export function build(data: any, opt: QueryConfig = {}): Schema {
...
   /** @return a list of the field names (for enumerating). */
-  public fields() {
-    return this._fieldSchemas.map((fieldSchema) => fieldSchema.field);
+  public names() {
+    return this._fieldSchemas.map((fieldSchema) => fieldSchema.name);
   }
 
   /** @return a list of FieldSchemas */
-  public get fieldSchemas() {
+  public get fields() {
```
vlandham: The `TableSchema` spec requires an array called `fields` - so thats why i changed it:

https://specs.frictionlessdata.io/table-schema/#descriptor

src/schema.ts
```diff
@@ -131,44 +164,44 @@ export function build(data: any, opt: QueryConfig = {}): Schema {
   return new Schema(fieldSchemas);
 }
 
-export class Schema {
+export class Schema implements TableSchema {
```
kanitw: Also, when I try

```
interface AI {
  a: string;
}

class A implements AI {
  
}
```
I got an error that `A` doesn't have `a: string`. 

I'm not sure why there is no similar error for schema here.

vlandham: I believe its because `Schema` does indeed implement the required field of `TableSchema` - which is `fields`

src/schema.ts
kanitw: Now I wonder why we can override `fields` array as a getter.  This seems to cloak the `fields` that comes with TableSchema.  

Given this and [the comment here](https://github.com/vega/compassql/pull/379/files#r121759269), I start wondering if we better make Schema encapsulate TableSchema rather than `implements` it as discussed [here](https://github.com/vega/compassql/pull/379/files#r121749686)?

src/schema.ts
kanitw: Ah! Just noticed that other properties are optional.

kanitw: But this means that we cannot access other properties of TableSchema via Schema anyway?

Here is a small code that I experiment with:
```
interface AI {
  a: string;
  b?: number;
}

class A implements AI {
  public a: string;
}

const aa = new A();
aa.a = 'a';
aa.b = 1; // error Property b does not exist on type A
```

vlandham: Right - cause we are not extending it - but indicating that Schema conforms to the interface. If there is optional values in the interface - we wouldn't expect to always have those optional values.

kanitw:
> This introduces some inconsistencies between this new Schema and (for example) the FieldQuery interface (field === name). we can work toward changing more interfaces if this is a problem.

We definitely should keep `field` in `FieldQuery` as it should be consistent with Vega-Lite.

src/schema.ts
kanitw: Yeah, but it's kinda weird that we don't make them accessible.  

Now I start liking encapsulation more again.

If `tableSchema: ReadOnly<TableSchema<FieldSchema>>` is a property of Schema, then we can easily export the TableSchema object.  In addition, we can directly access all of the TableSchema via Schema?

Note that Schema's constructor is only called via the `build()` method, so we have the freedom to modify it however we like.

src/schema.ts
```diff
@@ -27,6 +46,17 @@ export interface FieldSchema {
 }
 
 /**
+ * Table Schema
+ * see: https://specs.frictionlessdata.io/table-schema/
+ */
+export interface TableSchema {
```
vlandham: i think we have figured this out below - but just to clarify - `_fieldSchemas` is not redundant - it is just implementing the interface in a particular way.

src/constraint/field.ts
```diff
@@ -217,13 +217,13 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
       }
 
       switch (primitiveType) {
-        case PrimitiveType.BOOLEAN:
-        case PrimitiveType.STRING:
+        case DataType.BOOLEAN:
```
vlandham: okay

src/schema.ts
vlandham: in this scenario - what is the argument of TableSchema being parameterizable with `<FieldSchema>` ? In order to _be_ Table Schema, it needs to hold fields of a particular type. I believe the idea that `FieldSchema` **extends** `TableSchemaFieldDescriptor` would still work well in this encapsulated approach - and would simplify the signature of these interfaces.

src/schema.ts
kanitw: Ok now I understand how it works better you can ignore this comment if we chose to do `implements` rather than composition.  (If we do composition, then I still prefer the original name.)

src/schema.ts
kanitw: > in this scenario - what is the argument of TableSchema being parameterizable with <FieldSchema> ?

If we make it generic, we could do.  

```
interface TableSchema<F extends TableSchemaFieldDescriptor> {
   field: F;
   ...
}
```

kanitw:
cc: @AkshatSh   Note that this PR may cause some conflicts with your #372

vlandham:
@kanitw I believe this new commit addresses the issues you brought up.

* Now Schema contains a dataTableSchema - and uses it as the main schema storage.
* the `build()` function can now take a dataTable schema - which will serve as the basis of the dataTableSchema used by `Schema`. 
* Tests have been added to define the interaction between the existing inferring of types and the input dataTable (input dataTable trumps inferred types).
* primitiveType is back

src/schema.ts
```diff
@@ -40,36 +71,44 @@ export interface FieldSchema {
  *
  * @return a Schema object
  */
-export function build(data: any, opt: QueryConfig = {}): Schema {
+export function build(data: any, opt: QueryConfig = {}, tableSchema: TableSchema<TableSchemaFieldDescriptor> = {fields:[]}): Schema {
```
kanitw: Maybe opt should remain the last parameter?

src/schema.ts
```diff
@@ -8,25 +8,56 @@ import * as dlBin from 'datalib/src/bins/bins';
...
+
+/**
+ * Field Schema
+ */
+export interface FieldSchema extends TableSchemaFieldDescriptor {
+  vlType?: VLType;
 
-  title?: string;
   index?: number;
+  inIndex?: number;
```
kanitw: Maybe `originalIndex` is more descriptive?

src/schema.ts
```diff
@@ -128,47 +174,63 @@ export function build(data: any, opt: QueryConfig = {}): Schema {
     }
   }
 
-  return new Schema(fieldSchemas);
+  const derivedTableSchema: TableSchema<FieldSchema> = {
```
kanitw:
```
const derivedTableSchema: TableSchema<FieldSchema> = {
   ...tableSchema,
   fields: fieldSchemas
};
```
would be shorter and also support future additional properties in the table schema?

src/schema.ts
```diff
@@ -78,17 +117,24 @@ export function build(data: any, opt: QueryConfig = {}): Schema {
...
     } else {
-      type = Type.NOMINAL;
+      vlType = VLType.NOMINAL;
     }
 
-    return {
-      field: field,
+    let fieldSchema = {
+      name: name,
+      inIndex: inIndex,
```
kanitw: Add `// Need to keep original index for re-exporting TableSchema`

src/schema.ts
```diff
@@ -40,36 +71,44 @@ export interface FieldSchema {
...
-    let primitiveType: PrimitiveType = types[field] as any;
+  let tableSchemaFieldIndex = tableSchema.fields.reduce((m, field: TableSchemaFieldDescriptor) => {
+    m[field.name] = field;
+    return m;
+  }, {});
+
+  let fieldSchemas: FieldSchema[] = summaries.map(function(fieldProfile, index) {
+
+    const name: string = fieldProfile.field;
+    const inIndex: number = index;
```
kanitw: Since inIndex is only used once below in `let fieldSchema` -- just refer to `index` directly there?

src/schema.ts
kanitw: (`inIndex` confused me for a while)

vlandham:
@kanitw good suggestions. i believe i cover them all in the latest commit.

kanitw:
Thanks! :)

PR #384, opened 2017-06-21
Fix #186: Prevent occlusion by aggregating non positional fields

### LINKED ISSUE #186: Take mark properties channel into account for omitTableWithOcclusionIfAutoAddCount
https://github.com/uwdata/compassql/pull/182/files#diff-795bef163e2991324be9c36753d7fa4aR511

Basically, if there are detail/color/shape/ etc.  There are still occlusion.  Therefore, the existing constraints are not sufficient.

kanitw:


src/constraint/spec.ts
```diff
@@ -719,12 +719,27 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
-        ) {
-          return false;
+            (!isFieldQuery(yEncQ) || isDiscrete(yEncQ))) {
+          if (!specM.isAggregate()) {
+            return false;
+          } else {
+            return every(specM.getEncodings(), (encQ) => {
+              let channel = encQ.channel;
+
+              if (channel === Channel.COLOR || channel === Channel.DETAIL ||
```
kanitw: Instead checking for all non-xy channel, check if they are not x/y/row/column. 

(The number of non-position can grow in the future and manually checking one by one like this will make this constraint easily outdated.)

src/constraint/spec.ts
```diff
@@ -719,12 +719,27 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+                  channel === Channel.SHAPE || channel === Channel.SIZE ||
+                  channel === Channel.OPACITY) {
+                if (encQ['aggregate']) {
```
kanitw: `['aggregate']` is a hack to get around type system.  Instead, please use the new `isFieldQuery` or `isAutoCountQuery` type guard functions.

test/constraint/spec.test.ts
```diff
@@ -1415,6 +1415,55 @@ describe('constraints/spec', () => {
...
+          });
+          assert.isFalse(SPEC_CONSTRAINT_INDEX['omitTableWithOcclusionIfAutoAddCount'].satisfy(specM, schema, {autoAddCount: true}));
+        });
+      });
+    });
+
+    it('return false for plot with both x and y as dimension where any non-X, non-Y,' +
+       'non-Row, or non-Column channel do not have autoCount and is not aggregated', () => {
+      [Mark.POINT, Mark.CIRCLE, Mark.SQUARE, Mark.LINE, Mark.AREA, Mark.BAR].forEach((mark) => {
+        [Channel.COLOR, Channel.DETAIL, Channel.SHAPE, Channel.SIZE, Channel.OPACITY].forEach((rawChannel) => {
```
kanitw: You can remove size from here and use size with mean below instead of opacity.

test/constraint/spec.test.ts
```diff
@@ -1415,6 +1415,55 @@ describe('constraints/spec', () => {
...
+            {channel: Channel.X, field: 'N', type: Type.NOMINAL},
+            {channel: Channel.Y, field: 'N20', type: Type.NOMINAL},
+            {aggregate: 'mean', channel: Channel.SIZE, field: 'Q', type: Type.QUANTITATIVE}
+          ]
+        });
+        assert.isTrue(SPEC_CONSTRAINT_INDEX['omitTableWithOcclusionIfAutoAddCount'].satisfy(specM, schema, {autoAddCount: true}));
+      });
+    });
+
+    it('return false for plot with both x and y as dimensions where any non-X, non-Y' +
```
kanitw: Can you replace "non-X, non-Y, non-Row, or non-Column" with "non-position/facet" here and below?

felixcodes:
Fix #186

PR #389, opened 2017-06-22
Fix #85: Do not apply bin to a quantitative field with too low cardinality

### LINKED ISSUE #85: Do not apply bin to a quantitative field with too low cardinality

felixcodes:
Fix #85 
Min cardinality threshold is currently set at 15, but we can move it up or down later after playing with it.

Please:
- [x] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
- [x] Provide a test case
- [x] Make lint and test pass. (Run `npm run lint` and `npm run test`)
- [x] Make sure you have merged `master` into your branch.
- [x] Provide a concise title so that we can just copy it to our release note.

### DIFF
--- a/src/constraint/field.ts
+++ b/src/constraint/field.ts
@@ -28,7 +28,6 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
       // TODO: some aggregate function are actually supported by ordinal
       return true; // no aggregate is okay with any type.
     }
-  // TODO: minCardinalityForBin
   },{
     name: 'asteriskFieldWithCountOnly',
     description: 'Field="*" should be disallowed except aggregate="count"',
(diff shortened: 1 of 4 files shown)


src/constraint/field.ts
```diff
@@ -38,7 +37,22 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
...
+  },{
+    name: 'minCardinalityForBin',
+    description: 'binned quantitative field should not have too low cardinality',
+    properties: [Property.BIN, Property.FIELD, Property.TYPE],
+    allowWildcardForProperties: false,
+    strict: true,
+    satisfy: (fieldQ: FieldQuery, schema: Schema, _: PropIndex<Wildcard<any>>, opt: QueryConfig) => {
+      if (fieldQ.bin && fieldQ.type === Type.QUANTITATIVE) {
+        // We remove bin so schema can infer the raw unbinned cardinality.
+        // Q: Is this OK, or should I instead duplicate fieldQ and use delete operator
```
kanitw: This is fine.   If you want the whole fieldQ, you could do this (which is super short). 
```
const {bin: _, ...nonBinFieldQ} = fieldQ;
```

PR #392, opened 2017-06-27
Allow loading schema without overriding

### LINKED ISSUE #385: Allow loading schema without overriding
We made sure that importing and exporting TableSchema fields will make sure the output does not contain less information. 

However, given that users can modify some information, exporting and importing the same schema file may override user specified information.

cc: @vlandham

felixcodes:
Fix #385: Move field schema sorting into Schema constructor. 

Schema constructor now sorts field schemas, including field schemas from previously exported table schema. Note that exported table schemas are un-sorted back to the original order they had prior to build method due to table schema requirements. Moving sorting from `build` to Schema constructor allows field schema order to be consistent whether it was created by the Schema constructor or `build` factory method.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -145,29 +145,6 @@ export function build(data: any,  tableSchema: TableSchema<TableSchemaFieldDescr
     return fieldSchema;
   });
 
-  // order the fieldSchemas (sort them)
-  const order = {
-    'nominal': 0,
-    'key': 1,
-    'ordinal': 2,
-    'temporal': 3,
-    'quantitative': 4
-  };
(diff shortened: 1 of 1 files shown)


src/schema.ts
```diff
@@ -197,6 +174,32 @@ export class Schema {
 
   constructor(tableSchema: TableSchema<FieldSchema>) {
     this._tableSchema = tableSchema;
+
+    // order the field schema when we construct a new Schema
+    // this orders the fields in the UI
+    const order = {
```
kanitw: I think we should just move this `const` outside.  :)

PR #394, opened 2017-06-28
Add Domain to Schema for customizing ordinal's natural order

### LINKED ISSUE #87: Add Domain to Schema for customizing ordinal's natural order
- [x] Add optional `ordinalDomain` property to the `FieldSchema` for specifying natural order for ordinal data.  

- [x] make sure schema `ordinalDomain` method return this #131

- [x] Stick this `ordinalDomain` object to the output `scale.domain` in `SpecQueryModel.toSpec()`
for ordinal fields -- except when the query explicit have `domain` already.  (Do this in generate.ts -- after enumerating all properties)
  - [x] make sure to assign only if `ordinalDomain` exists

## Voyager 

- [x] action `DATASET_SCHEMA_CHANGE_ORDINAL_DOMAIN`

felixcodes:
Part of #87: 

- Add an optional `ordinalDomain` property to the `FieldSchema` for specifying natural order for ordinal data.
- Make `Schema`'s domain() method return `ordinalDomain` when applicable
- Make `SpecQueryMoel.toSpec()` output the value of `ordinalDomain` when applicable

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -292,7 +292,7 @@ export class SpecQueryModel {
         // if the property is a wildcard, return null
         if (isWildcard(encQ[prop])) return null;
 
-        // otherwise, assign the proper to the field def
+        // otherwise, assign the property to the field def
         if (encQ[prop] !== undefined) {
 
           if (!PROPERTY_SUPPORTED_CHANNELS[prop] ||  // all channels support this prop
@@ -305,8 +305,21 @@ export class SpecQueryModel {
 
                 let fieldSchema = this._schema.fieldSchema(field as string);
                 if (fieldSchema.ordinalDomain && !encQ[prop]['domain']) {
-                  fieldDef[prop] = extend(encQ[prop], {domain: fieldSchema.originalIndex});
+                  fieldDef[prop] = extend(encQ[prop], {domain: fieldSchema.ordinalDomain});
                 }
+              }
+            }
+          }
+          // if the encoding query doesn't have scale, but the schema does
+        } else if (prop === Property.SCALE) {
+          if (!PROPERTY_SUPPORTED_CHANNELS[Property.SCALE] ||
+            PROPERTY_SUPPORTED_CHANNELS[Property.SCALE][encQ.channel as Channel]) {
+            if (isFieldQuery(encQ)) {
+              let field = encQ.field;
+
+              let fieldSchema = this._schema.fieldSchema(field as string);
+              if (fieldSchema.ordinalDomain) {
+                fieldDef[Property.SCALE] = {domain: fieldSchema.ordinalDomain};
               }
             }
           }
--- a/src/model.ts
+++ b/src/model.ts
@@ -298,6 +298,17 @@ export class SpecQueryModel {
           if (!PROPERTY_SUPPORTED_CHANNELS[prop] ||  // all channels support this prop
             PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel]) {
             fieldDef[prop] = encQ[prop];
+
+            if (prop === Property.SCALE) {
+              if (isFieldQuery(encQ)) {
+                let field = encQ.field;
+
+                let fieldSchema = this._schema.fieldSchema(field as string);
+                if (fieldSchema.ordinalDomain && !encQ[prop]['domain']) {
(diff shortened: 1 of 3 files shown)


src/model.ts
```diff
@@ -298,6 +298,17 @@ export class SpecQueryModel {
           if (!PROPERTY_SUPPORTED_CHANNELS[prop] ||  // all channels support this prop
             PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel]) {
             fieldDef[prop] = encQ[prop];
+
+            if (prop === Property.SCALE) {
+              if (isFieldQuery(encQ)) {
+                let field = encQ['field'];
```
kanitw: Please use `.field` and never use string accessor (e.g., `['field']`) to hack around type except inside type guard.  (If you have to hack like this, it's likely that something is wrong.)

src/model.ts
```diff
@@ -292,12 +292,36 @@ export class SpecQueryModel {
...
 
           if (!PROPERTY_SUPPORTED_CHANNELS[prop] ||  // all channels support this prop
             PROPERTY_SUPPORTED_CHANNELS[prop][encQ.channel as Channel]) {
             fieldDef[prop] = encQ[prop];
+
+            if (prop === Property.SCALE) {
+              if (isFieldQuery(encQ)) {
+                let field = encQ.field;
+
+                let fieldSchema = this._schema.fieldSchema(field as string);
```
kanitw: - Check if it is ordinal first  and use `schema.domain({field, type})`

- Modify `schema.domain()` to return ordinalDomain if `type` is `"ordinal"` and `ordinalDomain` exists.

felixcodes: Note: We decided it's OK to read directly from `fieldSchema` since we are only assigning scale domain to fieldDef if the domain came from `fieldSchema.ordinalDomain`. A call to Schema's domain() accessor method cannot tell whether that domain came from `ordinalDomain` or `keys(fieldSchema.stats.unique);`

src/model.ts
```diff
@@ -290,14 +290,39 @@ export class SpecQueryModel {
...
 
         // if the property is a wildcard, return null
-        if (isWildcard(encQ[prop])) return null;
-
-        // otherwise, assign the proper to the field def
-        if (encQ[prop] !== undefined) {
+        if (isWildcard(encQ[prop])) {
+          return null;
+        } else if (encQ[prop] !== undefined) {
+        // otherwise, assign the property to the field def
```
kanitw: tab the comment

src/model.ts
```diff
@@ -290,14 +290,39 @@ export class SpecQueryModel {
...
             fieldDef[prop] = encQ[prop];
+
+            if (prop === Property.SCALE) {
+              if (isFieldQuery(encQ) && encQ.type === Type.ORDINAL) {
+                let field = encQ.field;
+
+                let fieldSchema = this._schema.fieldSchema(field as string);
+                // we use ['domain'] accessor hack because Typescript can't infer between FlatQuery and Wildcard
+                if (fieldSchema.ordinalDomain && !encQ[prop]['domain']) {
+                  fieldDef[prop] = extend(encQ[prop], {domain: fieldSchema.ordinalDomain});
```
kanitw: use 

```
fieldDef[prop] = {
   ... encQ[prop], 
  domain: fieldSchema.ordinalDomain
};
```

to avoid side effect to encQ and improve readability.

This would be equivalent to:
```
fieldDef[prop] = extend({} encQ[prop], {domain: fieldSchema.ordinalDomain});
```

src/model.ts
```diff
@@ -290,14 +290,39 @@ export class SpecQueryModel {
...
+                }
+              }
+            }
+          }
+          // if the encoding query doesn't have scale, but the schema does
+        } else if (prop === Property.SCALE) {
+          if (!PROPERTY_SUPPORTED_CHANNELS[Property.SCALE] ||
+            PROPERTY_SUPPORTED_CHANNELS[Property.SCALE][encQ.channel as Channel]) {
+            if (isFieldQuery(encQ) && encQ.type === Type.ORDINAL) {
```
kanitw: You are repeating your code twice.  I'm sure you can you better logic that wrap this chunk to avoid duplicate code.

kanitw: (This would be a good exercise that is similar to coding interview -- they will ask you to improve your code :p)

PR #405, opened 2017-09-06
Export missing Query type

herrstucki:
So one can write in TS

```ts
const query: cql.query.Query = {...}
```

… instead of having to do this:

```ts
import { Query } from "compassql/build/src/query/query";
```


kanitw:
Thanks for the PR.  Btw, just curious what you're using CompassQL for? ;)

PR #409, opened 2017-10-17
add support for rect mark

haldenl:
Support rect markings in search space.


src/constraint/spec.ts
```diff
@@ -657,7 +659,15 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
             }
             return false;
           }
-
+        case Mark.RECT:
+          // Until CompassQL supports layering, it only makes sense for
+          // rect to encode DxD (otherwise just use bar).
+          const xEncQ = specM.getEncodingQueryByChannel(Channel.X);
+          const yEncQ = specM.getEncodingQueryByChannel(Channel.Y);
+          const xIsDimension = isDimension(xEncQ);
```
kanitw: It doesn't always have to be a dimension (1D) works too.

haldenl:
I made a few changes:
- added a test for DxDxQ in generate.test.ts (not sure if this is the correct place to put it, recommend.test.ts looks similarly sparse).
- adding support for 1xD + color rect encoding without generating bad results required a special case in the 'omitNonPositionalOrFacetOverPositionalChannels' and adding another constraint 'omitFacetOverPositionalChannels'. They seem a little redundant together, I was wondering if it would make sense to split 'omitNonPositionalOrFacetOverPositionalChannels' into two separate tests, one for NonPositional and another for Facet.

src/ranking/effectiveness/mark.ts
```diff
@@ -262,8 +262,8 @@ function init() {
     DISCRETE_OR_NONE.forEach((yType) => {
       // has occlusion
       const ddMark = {
-        point: 0,
         rect: 0,
+        point: -0.0025,
```
kanitw: I wonder if there are corner cases where making point inferior to rect in this case cause a bit more problem.

test/generate.test.ts
```diff
@@ -299,6 +299,35 @@ describe('generate', function () {
...
+        },{
+          channel: SHORT_WILDCARD,
+          field: 'Q',
+          type: Type.QUANTITATIVE
+        }]
+      };
+
+      const answerSet = generate(specQ, schema, CONFIG_WITH_AUTO_ADD_COUNT);
+      it('should generate a rect heatmap with color encoding the quantitative measure', () => {
+        answerSet.forEach((specM) => {
```
kanitw: Currently the test checks if the result contains rect with color. 
We may want to write another test that make sure we recommend rect with color over other cases.  

(This test is good, but just we may want to add more test.)

test/ranking/effectiveness/mark.test.ts
```diff
@@ -217,7 +217,7 @@ export const DD_RULESET = {
     const _rules: Rule[] = [];
     [TIMEUNIT_O, O, BIN_Q, N].forEach((xType) => {
       [TIMEUNIT_O, O, BIN_Q, N].forEach((yType) => {
-        const ddRank = [POINT, TEXTMARK, TICK, [BAR, LINE, AREA], RULE];
+        const ddRank = [RECT, POINT, TEXTMARK, TICK, [BAR, LINE, AREA], RULE];
```
kanitw: REMOVE the todo below about rect

src/constraint/spec.ts
```diff
@@ -429,37 +430,64 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
-        } else if (channel === Channel.Y) {
-          hasY = true;
-        } else if (!isWildcard(channel)) {
-          // All non positional channel / Facet
-          hasNonPositionalChannelOrFacet = true;
-          if (specM.wildcardIndex.hasEncodingProperty(i, Property.CHANNEL)) {
-            hasEnumeratedNonPositionOrFacetChannel = true;
+      switch(mark) {
+        case Mark.RECT:
+          // RECT is a special exception, as it can be used in a 1xD manner
```
kanitw: I'm not sure if this constraint should be relaxed for rect?

If there is rect 1D with color, it is still better to use bar 2D (move color to another position channel) anyway?

src/constraint/spec.ts
```diff
@@ -657,7 +685,26 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+          // rect to encode DxD or 1xD (otherwise just use bar).
+          // Furthermore, color should only be used in a 'heatmap' fashion
+          // (with a measure field).
+          const xEncQ = specM.getEncodingQueryByChannel(Channel.X);
+          const yEncQ = specM.getEncodingQueryByChannel(Channel.Y);
+          const xIsDimension = isDimension(xEncQ);
+          const yIsDimension = isDimension(yEncQ);
+
+          const colorEncQ = specM.getEncodingQueryByChannel(Channel.COLOR);
+          const colorIsQuantitative = isMeasure(colorEncQ);
```
kanitw: What if color is ordinal?   (In the case where the data have no duplicate pairs of x, y values.)

src/constraint/spec.ts
```diff
@@ -429,37 +430,64 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
...
+              (opt.constraintManuallySpecifiedValue && hasNonPositionalChannelOrFacet)
+            ) {
+            return hasX && hasY;
+          }
+          return true;
       }
+    }
+  },
+  {
+    name: 'omitFacetOverPositionalChannels',
```
kanitw: I still do not understand why you need to split this constraint from the other.  Could you please explain more?

test/generate.test.ts
kanitw: Perhaps, we may need to add more test to the ranking part?

src/constraint/spec.ts
haldenl: If I recall correctly, in one of our previous conversations you mentioned this 1D w/ color should be a possibility (and that's why I relaxed the constraint). I think it does make more sense to just use a bar in that case, however.

src/constraint/spec.ts
haldenl: The split was motivated by 1D rect support. The original constraint also meant 'color' could not be used unless both positional channels had been filled (which would eliminate 1D rect), hence the exception for rect in that constraint. Without this second constraint, rows or columns appeared even for 1D rect encodings (e.g. multiple 1D rect encodings lined up). My thinking was that rect should be exempt from the first constraint, but all encodings should match the second constraint. It is a little redundant, hence my comment about splitting into just 'nonpositional' (maybe there is a better word for this that just encompasses color / size but excludes 'row' / 'column') and 'facet'.

src/constraint/spec.ts
haldenl: done!

test/ranking/effectiveness/mark.test.ts
haldenl: done

src/ranking/effectiveness/mark.ts
haldenl: Since rect should only compete with point in a DxDx(Q|O) query, there should only be two cases right? (seen below).

screen shot 2017-11-19 at 4 07 09 pm
screen shot 2017-11-19 at 4 54 31 pm

test/generate.test.ts
haldenl: Would these go in `mark.test.ts`? I see a Discrete-Discrete test there (where Rect > Point) but I'm not clear on how the tests in that file are able to handle 3rd dimensions like (with color Rect should be > Point but if the 3rd channel is unspecified then size is better and Point > Rect).

src/ranking/effectiveness/mark.ts
haldenl: Ah another case where we specific color and therefore Rect should beat Point

image

kanitw:
Not sure why there are so many new "result" files in this PR.  Could you please squash and rebase to the latest master to make this contain only relevant diff?  That will make it easier for me to review. (so I can focus on the core part of this PR.)

src/ranking/effectiveness/mark.ts
kanitw: If there is x:D,y:D,?:Q where Q is not aggregated, then perhaps points are better because at least users can notice the occlusion. 

If there is ?:D,?:D,?:Q  where Q is not aggregated, perhaps, having one of the Q on x or y maybe better?

We should also consider the cases for Q is aggregated -- that case is actually less problematic.

kanitw:
@haldenl  -- FYI, I've rebased this branch on top of https://github.com/vega/compassql/pull/422 and force pushed.   Make sure to remove your local branch + checkout the new one before making further changes.

src/constraint/spec.ts
kanitw: >  I think it does make more sense to just use a bar in that case, however.

I see.  I agree that bar still makes more sense anyway. 
Since this is a non-strict constraint, maybe we don't have to create a special case for rect?


In a way, point with x and size/color is a fair game too.  It's just that scatterplot is almost always more reasonable.  So it seems a bit inconsistent to relax this constraint only for rect.

haldenl: That sounds fair, I will remove this case

haldenl:
I don't believe it makes sense to split the constraint if we do not want to relax the first constraint for 1D rect. I will remove the second constraint!

src/ranking/effectiveness/mark.ts
haldenl: For now assume occlusion heuristics are correct and modify rankings accordingly

src/config.ts
```diff
@@ -45,6 +45,7 @@ export interface QueryConfig {
   omitRawWithXYBothOrdinalScaleOrBin?: boolean;
   omitRepeatedField?: boolean;
   omitNonPositionalOrFacetOverPositionalChannels?: boolean;
+  omitFacetOverPositionalChannels?: boolean;
```
kanitw: Do we still need to add this config?

src/ranking/effectiveness/mark.ts
haldenl: DxD Point + size >  DxD Rect + color > DxD Point + color

- Add a feature color channel to correct this.

- Because point > rect and point probably get 0; rect will get small minus (-R).  the rect x color can be a +R  and maybe do something small with point and color (minus  < |R|)

haldenl:
This should be good to go. To summarize the example changes, we see RECT used whenever we see POINT used for Dx1, DxD, or DxDxQ, with arbitrary ordering between the two. Unless the query specifies RECT mark, the highest ranked recommendation based on effectiveness will be POINT + SIZE over RECT + COLOR.

test/ranking/effectiveness/mark.test.ts
```diff
@@ -217,20 +217,19 @@ export const DD_RULESET = {
     const _rules: Rule[] = [];
     [TIMEUNIT_O, O, BIN_Q, N].forEach((xType) => {
       [TIMEUNIT_O, O, BIN_Q, N].forEach((yType) => {
-        const ddRank = [POINT, TEXTMARK, TICK, [BAR, LINE, AREA], RULE];
-
-        // TODO: RECT is not bad here.
-
+        const base = [TEXTMARK, TICK, [BAR, LINE, AREA], RULE];
```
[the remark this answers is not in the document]
kanitw: This is outdated ....

haldenl:
@kanitw do you have any ideas as to why the travis builds are failing? 

>Failed to install 5. Remote repository may not be reachable.

kanitw:
Not sure, but we can restart the build just in case it's due to server down.

Issue #410, opened 2017-10-17
Setup infrastructure to compile examples query and maintain output


kanitw:
- [x] have a script to run all examples with `recommend` method
- [x] each item would either a SpecQueryModel_Group_ or SpecQueryModel
  - [x] For specQueryModel, map them to shorthand ( `specQMs.map(x => x.toShorthand()` ) 
  - use `isSpecQueryGroup` to check if it is a group

- [x] check if the compiler has changed in Travis: 
  - look at `npm run build:examples-only` and then `npm run check:examples` in `.travis.yml` in Vega-Lite

kanitw:
@haldenl Is there any task left for this issue?

haldenl:
Everything is done!

PR #414, opened 2017-10-30
Add infrastructure for compiling cql examples to check against (integration testing)

haldenl:
Issue #410

Added scripts and modify configs to allow for building of output from compassQL examples, as well as automated example diff testing on commits.

`scripts/build-examples.sh` runs through all specs in `examples/specs` and gives them to `scripts/cql2shorthand.js`, which generates `${name}.results.json` output files, placing them in `examples/output`. Each of these output files is a json array of shorthands (see `src/query/shorthand.ts`) for the `SpecQuery`s recommended by the input compassQL spec). The order of these encodings follows a pre-order traversal of the resultant `SpecQueryModelGroup` of `recommend(...).result`. 

Output diff checking is also included for `travis` for additional integration testing.


scripts/cql2shorthand.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // options
+  const opt = {};
+
+  // get the recommendation
+  const output = cql.recommend(query, schema, opt);
+
+  // get the resulting collapsed specs
+  const results = [];
+  cql.result.collapseTree(output.result, (item) => {
```
kanitw: I think we should preserve the tree structure rather than doing a pre-order traversal.

haldenl:
Updated to preserve tree structure (and removed the collapse function)

PR #415, opened 2017-11-09
Add width/height as a top level specification for CQL queries.

haldenl:
Width height can be specified to control the dimensions of the output vega-lite specs that come out of `recommend`. See issue #332

### DIFF
--- a/src/query/shorthand.ts
+++ b/src/query/shorthand.ts
@@ -8,7 +8,7 @@ import {isString} from 'datalib/src/util';
 import {EncodingQuery, isFieldQuery, FieldQuery, isValueQuery, isDisabledAutoCountQuery, isEnabledAutoCountQuery, isAutoCountQuery, FieldQueryBase} from './encoding';
 import {SpecQuery, stack, fromSpec} from './spec';
 import {isWildcard, isShortWildcard, SHORT_WILDCARD} from '../wildcard';
-import {getEncodingNestedProp, Property, isEncodingNestedParent, DEFAULT_PROP_PRECEDENCE, SORT_PROPS, EncodingNestedChildProp} from '../property';
+import {getEncodingNestedProp, Property, isEncodingNestedParent, DEFAULT_PROP_PRECEDENCE, SIZE_PROPS, SORT_PROPS, EncodingNestedChildProp} from '../property';
 import {PropIndex} from '../propindex';
 import {Dict, keys, isArray, isBoolean} from '../util';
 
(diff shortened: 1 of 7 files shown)


src/query/spec.ts
```diff
@@ -28,11 +28,17 @@ export interface SpecQuery {
   /**
    * Array of encoding query mappings.
    * Note: Vega-Lite's `encoding` is an object whose keys are unique encoding channels.
-   * However, for CompassQL, the `channel` property of encoding query mappings can be wildcards. 
+   * However, for CompassQL, the `channel` property of encoding query mappings can be wildcards.
    * Thus the `encoding` object in Vega-Lite is flatten as the `encodings` array in CompassQL.
    */
   encodings: EncodingQuery[];
 
+  /**
```
kanitw: Remove the JSDoc comment or change it to normal comment (`// width and height//`? 

Note that this comment would be only applied to height when you hover over width variable. So if we have descriptive comment, better make them separate for width / height.  However, if the `width` property's JSDocs is simply "width", there is no point adding it.

kanitw: I think a good JSDocs note for these two variables are that `width` cannot be wildcard (yet).  Same for `height`.  

(I don't think we need to implement them as wildcard yet as there is no required use case.)

haldenl: done!

kanitw:
Still need to do this? 

> I think we should add this to shorthand somehow too. Maybe as |width=100|height=100 or something.

And maybe test if this really works in the editor, if it works, capture a screenshot then I think we should be good to go.

haldenl:
Oh yes, I missed that comment, sorry. Should be done! Here's a screenshot.
screen shot 2017-11-21 at 6 32 55 am

src/query/shorthand.ts
```diff
@@ -134,6 +134,13 @@ export function spec(specQ: SpecQuery,
     }
   }
 
+  if (!!specQ.width) {
```
kanitw: I think you have to take `include` into account here too.  

(I also notice that we forgot to do the same for `transform` above -- please help fix that too -- maybe in a separate PR.)

src/query/spec.ts
```diff
@@ -28,11 +28,21 @@ export interface SpecQuery {
...
+   * The width of the resulting encodings, does not support wildcards.
```
kanitw: I think splitting the sentence makes it a bit easier to read :) 

```
The width of the resulting encodings.

__Note:__ This property does not support wildcards.
```

kanitw:
One thing that I start wondering is whether we should add other top-level properties of to CompassQL.  (They are probably less important than adding `width` and `height`, but given that you already have context about how to add them, maybe it's worth spending a little more time to add other properties.) -- but let's do that in a separate PR.

kanitw:
Don't forget to take care of "include" :)

haldenl:
Refactored for include!

haldenl:
Okay sounds, good, would we want to include mark also as a 'Top Level Property'?

kanitw:
> Okay sounds, good, would we want to include mark also as a 'Top Level Property'?

Um let's say no -- because mark can have nested properties and top-level properties in VL is applicable for single view and multi view (while mark is only for single view)

kanitw:
I think your latest commit which says "add test" haven't really add test. lol

PR #417, opened 2017-11-21
Add more comments about `autoCount`

kanitw:


kanitw:
FYI, I don't care about wrapping comments within 80 or 100 characters. 

I mean it doesn't hurt to do, but given that our code review tools (Github) and our text editor can wrap text.  I don't see why we have to do it.  For code, it is definitely more important to wrap them.  

Thanks for doing it for me though :)

PR #427, opened 2018-01-16
Add support for remaining top level props

haldenl:
Add support for background, title, padding-top level props.
Still remaining: AutoSize, but requires vega-lite dependency update (for future PR).

These appear in shorthand as a top-level prop, with string values appearing directly, e.g.:
`background: "gray"`

 and object values appearing in JSON.stringify format, e.g:
`title: {"text": "A Simple Bar Chart", "anchor": "start"}`


kanitw:
Neat!

PR #442, opened 2018-07-04
Update package.json's main and other props to mirror VL format

ssharif6:
Change properties in `package.json` to mirror VL's `package.json` format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,11 @@
   "name": "compassql",
   "version": "0.20.1",
   "description": "CompassQL visualization query language",
-  "main": "build/src/cql.js",
+  "main": "build/compassql.js",
+  "unpkg": "build/compassql.min.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,11 @@
   "name": "compassql",
   "version": "0.20.1",
   "description": "CompassQL visualization query language",
-  "main": "build/src/cql.js",
+  "main": "build/compassql.js",
+  "unpkg": "build/compassql.min.js",
+  "jsdelivr": "build/compassql.min.js",
+  "module": "build/src/index",
```
kanitw: we don't have index.ts -- we still use the old school cql.ts as the main file.  Same for types.

kanitw: Let's correct this to cql.

kanitw:
Feefeel free to release

PR #444, opened 2018-07-12
Ignore ValueDef when performing scoring

ssharif6:
Add support for ValueDef

Ignore scoring for ValueQueries

### DIFF
--- a/src/ranking/effectiveness/typechannel.ts
+++ b/src/ranking/effectiveness/typechannel.ts
@@ -1,11 +1,10 @@
 
 
-import {Channel} from 'vega-lite/build/src/channel';
 
 import {QueryConfig} from '../../config';
 import {SpecQueryModel} from '../../model';
 import {fieldDef as fieldDefShorthand} from '../../query/shorthand';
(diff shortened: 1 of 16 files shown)


src/ranking/effectiveness/typechannel.ts
```diff
@@ -98,23 +98,27 @@ export class TypeChannelScorer extends Scorer {
 
   public getScore(specM: SpecQueryModel, schema: Schema, opt: QueryConfig): FeatureScore[] {
     const encodingQueryByField = specM.getEncodings().reduce((m, encQ) => {
-      const fieldKey = fieldDefShorthand(encQ);
-      (m[fieldKey] = m[fieldKey] || []).push(encQ);
-      return m;
+      if (isFieldQuery(encQ)) {
+        const fieldKey = fieldDefShorthand(encQ);
+        (m[fieldKey] = m[fieldKey] || []).push(encQ);
+        return m;
```
kanitw: Actually, you have to always `return m` in reduce.

PR #450, opened 2019-12-10
fix: move typescript-json-schema to devDependencies

oigewan:
### Issue
Installing this package ends up installing typescript version 2.x due to `typescript-json-schema` being listed as a dependency instead of a devDependency.  

### Changes
Move `typescript-json-schema` into the devDependencies section of package.json.

### Testing
`yarn run build && yarn run test` succeeded.


oigewan:
@kanitw Do you have any idea when this change will be released?

PR #453, opened 2020-10-16
ci: switch to gh actions

domoritz:
@kanitw do the changes look reasonable? They come only from updating node, nothing else.


haldenl:
why did all the outputs change? Quick glance it looks like just a osrting change but seems strange.

domoritz:
Because Node 12 and 14 produce different results from Node 10.

domoritz:
Otherwise good @haldenl @kanitw ?

PR #456, opened 2020-12-02
existing packages fully updated

leibatt:
Please:
- [x] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
    - the changes could not be decoupled, unfortunately. This is a big change.
- [ ] Provide a test case (N/A?)
- [x] Make lint and test pass. (Run `npm run lint` and `npm run test`)
    - Note: still needs to be upgraded to eslint, but tests pass with tslint.
- [x] Make sure you have merged `master` into your branch.
- [x] Provide a concise title so that we can just copy it to our release note.
  - Use imperative mood and present tense.
  - Mention relevant issues. (e.g., `#1`)


src/schema.ts
```diff
@@ -3,8 +3,9 @@ import {inferAll} from 'datalib/src/import/type';
 import {summary} from 'datalib/src/stats';
 import {autoMaxBins} from 'vega-lite/build/src/bin';
 import {Channel} from 'vega-lite/build/src/channel';
-import {containsTimeUnit, convert, TimeUnit, TIMEUNIT_PARTS} from 'vega-lite/build/src/timeunit';
+import {TimeUnitParams,SingleTimeUnit,LocalSingleTimeUnit,isUTCTimeUnit,containsTimeUnit, TimeUnit, TIMEUNIT_PARTS} from 'vega-lite/build/src/timeunit';
```
domoritz, line 6: Unused import TimeUnitParams

jest.config.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.jsx?$": "babel-jest"
```
domoritz, line 6: Do we need to specify this? I thought jest uses the babel transform automatically (and babel supports typescript).

In Vega-Lite, we almost don't need to override ay configs: https://github.com/vega/vega-lite/blob/master/package.json#L141

package.json
```diff
@@ -50,57 +50,34 @@
...
     "concurrently": "^5.3.0",
-    "jest": "^23.6.0",
-    "rollup": "^2.32.1",
+    "jest": "^26.6.3",
+    "rollup": "^2.33.1",
     "rollup-plugin-sourcemaps": "^0.6.3",
     "source-map-support": "^0.5.19",
     "terser": "^5.3.8",
-    "ts-jest": "^23.10.5",
+    "ts-jest": "^26.4.4",
```
domoritz, line 70: I don't think we need this since we have babel.

src/constraint/field.ts
```diff
@@ -90,15 +90,15 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
...
+      const {compatible} = channelCompatibility(fieldDef, fieldQ.channel as ExtendedChannel);
 
       if (compatible) {
         return true;
       } else {
         // In VL, facet's field def must be discrete (O/N), but in CompassQL we can relax this a bit.
         const isFacet = fieldQ.channel === 'row' || fieldQ.channel === 'column';
 
-        if (isFacet && (isLocalSingleTimeUnit(fieldDef.timeUnit) || isUtcSingleTimeUnit(fieldDef.timeUnit))) {
+        if (isFacet && !!fieldDef.timeUnit && (isLocalSingleTimeUnit(fieldDef.timeUnit as string) || isUTCTimeUnit(fieldDef.timeUnit as string))) {
```
domoritz, line 101: This will break when time unit is not a string. I guess this cannot be the case since we control what we generate.

src/constraint/field.ts
```diff
@@ -243,7 +243,7 @@ export const FIELD_CONSTRAINTS: EncodingConstraintModel<FieldQuery>[] = [
               // ignore type and properties of wildcards
               continue;
             }
-            let isSupported = channelScalePropertyIncompatability(channel, scaleProp as keyof Scale) === undefined;
+            let isSupported = channelScalePropertyIncompatability(channel as Channel, scaleProp as keyof Scale) === undefined;
```
domoritz, line 246: Why do we need to force channel to be a Channel? Can we support ExtendedChannel in channelScalePropertyIncompatability?

src/constraint/spec.ts
```diff
@@ -232,7 +232,9 @@ export const SPEC_CONSTRAINTS: SpecConstraintModel[] = [
         // channel unspecified, no need to check
         if (isWildcard(encQ.channel)) return true;
 
-        return !!supportMark(encQ.channel, mark as Mark);
+        if (encQ.channel === 'row' || encQ.channel === 'column') return true;
+
+        return !!supportMark(encQ.channel as CHANNEL.Channel, mark as Mark);
```
domoritz, line 237: Why do we need to force CHANNEL.Channel? If you extract a channel variable, typescript should be able to figure out that channel isn't row or column anymore.

src/query/encoding.ts
```diff
@@ -325,5 +330,5 @@ export function scaleType(fieldQ: FieldQuery) {
     timeUnit: timeUnit as TimeUnit,
     bin: bin as BinParams
   };
-  return compileScaleType({type: scale.type}, channel, fieldDef, markType);
+  return compileScaleType({type: scale.type}, channel as Channel, fieldDef, markType);
```
domoritz, line 333: channel can't be row or column anymore. Why do we need to force the type now?

src/schema.ts
```diff
@@ -3,8 +3,9 @@ import {inferAll} from 'datalib/src/import/type';
...
 import * as TYPE from 'vega-lite/build/src/type';
+import * as VEGA_TIME from 'vega-time';
```
domoritz, line 8:
```suggestion
import * as vegaTime from 'vega-time';
```

test/stylize.test.ts
```diff
@@ -26,7 +27,8 @@ describe('stylize', () => {
 
       specM = smallRangeStepForHighCardinalityOrFacet(specM, schema, {}, DEFAULT_QUERY_CONFIG);
       assert.equal(
-        ((specM.getEncodingQueryByChannel(CHANNEL.Y) as FieldQuery).scale as ScaleQuery).rangeStep,
+        //((specM.getEncodingQueryByChannel(CHANNEL.Y) as FieldQuery).scale as ScaleQuery).rangeStep,
```
domoritz, line 30: Fix or delete commented out code.

src/constraint/field.ts
leibatt: I'm not sure how to take a Typescript type and turn it into a string to match the function definition of `isLocalSingleTimeUnit` from Vega-Lite. It seems that maybe there should be a separate set of functions to take TimeUnit as input rather than strings as input in Vega-Lite's time unit functionality?

src/constraint/spec.ts
leibatt: I get a build error saying essentially that ExtendedChannel !== Channel when I do not cast the type of encQ.channel to CHANNEL.Channel. I changed encQ.channel to be of type ExtendedChannel to accommodate faceting.

src/query/encoding.ts
leibatt: `channel` is of type ExtendedChannel rather than of type `Channel`, which clashes with the expected type of Channel. Should I change the variable names to be extendedChannel instead of channel?

domoritz: Vega-Lite used to only have Channel but now we distinguish between Channel and ExtendedChannel (Channel + row, column, facet). I think CompassQL could support extended channels unless it doesn't make sense for the particular functionality.

What do you think?

src/constraint/spec.ts
domoritz: Then supportMark should support extended channel as well. If we just cast here, we may be passing an extended channel to something that supports only channel, which will cause bugs.

src/constraint/field.ts
domoritz: Can you normalize the time unit with `normalizeTimeUnit(timeUnit)?.unit`? We do something like that in https://github.com/vega/vega-lite/blob/master/src/channeldef.ts#L1322. 

isLocalSingleTimeUnit only supports strings.

leibatt: Yes it works thank you!

src/query/encoding.ts
leibatt: ExtendedChannels make the most sense to me for Compassql. The issue I'm having is that I am not sure how to adapt the code to work with the vl functions that expect Channels. It seems that if we've ruled out row, column, and facet as possibilities, an ExtendedChannel just becomes a Channel. Can TypeScript figure this out automatically? Am I just missing some checks in my code?

